PR #1, opened 2016-08-09
FreeBSD support

bukinr:


aswaterman:
looks good to me, @palmer-dabbelt can you give your approval & merge?

palmer-dabbelt:
@bukinr: have you done copyright assignment with the FSF?  If not and these headers are just copies from somewhere else (which they seem to be), then it should be OK.  Otherwise you'll need to do copyright assignment so we can upstream this, I can walk you through things via email if you need help.

aswaterman:
We'd obviously like FreeBSD support in GCC, but we're in the process of submitting patches now.  Since we didn't hear back, I'm going to close this PR.  If you do have copyright assignment on file, feel free to reopen in the next 24 hours or so; otherwise, the patches should just go to GCC directly post-upstreaming.

PR #2, opened 2016-08-10
Resolve several gcc testsuite issue.

kito-cheng:
This patch set resolve about ~50% of gcc testsuite fail.

configure: riscv64-unknown-elf + soft-float

gcc testsuite result before those patch:

                === gcc Summary ===

 # of expected passes            79639
 # of unexpected failures        450
 # of unexpected successes       1
 # of expected failures          101
 # of unresolved testcases       143
 # of unsupported tests          1653

                === g++ Summary ===

 # of expected passes            82555
 # of unexpected failures        1242
 # of expected failures          260
 # of unresolved testcases       5
 # of unsupported tests          3960

gcc testsuite result after those patch with gdb sim fix[1]:
                === gcc Summary ===

 # of expected passes            79827
 # of unexpected failures        214 (-236)
 # of unexpected successes       1
 # of expected failures          102
 # of unresolved testcases       71
 # of unsupported tests          1664

                === g++ Summary ===
 # of expected passes            82578
 # of unexpected failures        1222 (-20)
 # of expected failures          260
 # of unresolved testcases       2
 # of unsupported tests          3960

[1] https://github.com/riscv/riscv-binutils-gdb/pull/15

### DIFF
--- a/gcc/config/riscv/riscv.h
+++ b/gcc/config/riscv/riscv.h
@@ -702,7 +702,7 @@ enum reg_class
 /* Accept arguments in a0-a7 and/or fa0-fa7. */
 #define FUNCTION_ARG_REGNO_P(N)					\
   (IN_RANGE((N), GP_ARG_FIRST, GP_ARG_LAST)			\
-   || IN_RANGE((N), FP_ARG_FIRST, FP_ARG_LAST))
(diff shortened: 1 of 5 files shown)


aswaterman:
awesome!!

PR #4, opened 2016-10-05
Fix multi-lib for RV32I

kito-cheng:
Don't use multi-lib option to soft-float when it's default since it's will make gcc can't find library in default path (eg: lib or lib32).


aswaterman:
Fine with me, @palmer-dabbelt ?

palmer-dabbelt:
Thanks!

palmer-dabbelt:
This was merged manually.

PR #7, opened 2016-10-13
add options to individually include hw mul and div/mod

vanjoe:
I added -mmul and -mdiv to the target specific options. I kept -mmuldiv around for legacy. 

If anyone knows a better way of making -mmuldiv simply expand to -mmul -mdiv, I'd like to know, otherwise i think my way works.

### DIFF
--- a/gcc/common/config/riscv/riscv-common.c
+++ b/gcc/common/config/riscv/riscv-common.c
@@ -45,9 +45,9 @@ riscv_parse_arch_string (const char *isa, int *flags)
       return;
     }
 
-  *flags &= ~MASK_MULDIV;
+  *flags &= ~(MASK_MUL | MASK_DIV);
   if (*p == 'M')
-    *flags |= MASK_MULDIV, p++;
+    *flags |= (MASK_MUL | MASK_DIV), p++;
 
   *flags &= ~MASK_ATOMIC;
   if (*p == 'A')
@@ -110,7 +110,16 @@ riscv_handle_option (struct gcc_options *opts,
     case OPT_march_:
       riscv_parse_arch_string (decoded->arg, &opts->x_target_flags);
       return true;
-
+    case OPT_mmuldiv:
+      if (decoded->value)
+         {
+            opts->x_target_flags |= (MASK_MUL | MASK_DIV);
+         }
+      else
+         {
+            opts->x_target_flags &= ~(MASK_MUL | MASK_DIV);
+         }
+      return true;
     default:
       return true;
     }
--- a/gcc/config/riscv/riscv.h
+++ b/gcc/config/riscv/riscv.h
@@ -63,7 +63,11 @@ along with GCC; see the file COPYING3.  If not see
       if (TARGET_ATOMIC)						\
 	builtin_define ("__riscv_atomic");				\
 									\
-      if (TARGET_MULDIV)						\
+      if (TARGET_MUL)							\
+	builtin_define ("__riscv_mul");					\
+      if (TARGET_DIV)							\
+	builtin_define ("__riscv_div");					\
+      if (TARGET_DIV && TARGET_MUL)					\
 	builtin_define ("__riscv_muldiv");				\
 									\
       if (TARGET_HARD_FLOAT_ABI)					\
--- a/gcc/config/riscv/riscv.md
+++ b/gcc/config/riscv/riscv.md
@@ -566,13 +566,13 @@
   [(set (match_operand:GPR 0 "register_operand")
 	(mult:GPR (match_operand:GPR 1 "reg_or_0_operand")
 		   (match_operand:GPR 2 "register_operand")))]
-  "TARGET_MULDIV")
+  "TARGET_MUL")
 
 (define_insn "*mulsi3"
   [(set (match_operand:SI 0 "register_operand" "=r")
 	(mult:SI (match_operand:GPR 1 "register_operand" "r")
 		  (match_operand:GPR2 2 "register_operand" "r")))]
-  "TARGET_MULDIV"
+  "TARGET_MUL"
   { return TARGET_64BIT ? "mulw\t%0,%1,%2" : "mul\t%0,%1,%2"; }
   [(set_attr "type" "imul")
    (set_attr "mode" "SI")])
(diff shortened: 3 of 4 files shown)


gcc/common/config/riscv/riscv-common.c
```diff
@@ -110,9 +110,16 @@ riscv_handle_option (struct gcc_options *opts,
...
+	    opts->x_target_flags |= (MASK_MUL | MASK_DIV);
+	    //error("OPT_mmuldiv found");
+	    return true;
+    case OPT_mno_muldiv:
+	    opts->x_target_flags &= ~(MASK_MUL | MASK_DIV);
+	    //error("OPT_mno_muldiv found");
+	    return true;
     default:
-      return true;
+	    return true;
```
kito-cheng: Wrong indent and unclean up code.

gcc/config/riscv/riscv.opt
```diff
@@ -71,9 +71,21 @@ Target Report Mask(ATOMIC)
 Use hardware atomic memory instructions.
 
 mmuldiv
-Target Report Mask(MULDIV)
+Target Report RejectNegative Var(TARGET_muldiv)
 Use hardware instructions for integer multiplication and division.
 
+mno-muldiv
+Target Report RejectNegative Undocumented Var(TARGET_no_muldiv)
+Dont Use hardware instructions for integer multiplication and division.
```
kito-cheng: I think let remove  RejectNegative for mmuldiv is more reasonable since mno-muldiv is Undocumented, 
and Var(TARGET_muldiv) can remove if it's unused.

gcc/config/riscv/riscv.md
```diff
@@ -169,7 +169,7 @@
 ;; D2S          double to float single
 ;; S2D          float single to double
 
-(define_attr "cnv_mode" "unknown,I2S,I2D,S2I,D2I,D2S,S2D" 
+(define_attr "cnv_mode" "unknown,I2S,I2D,S2I,D2I,D2S,S2D"
```
kito-cheng: Clean up should split into another patch instead of mix with feature implementation patch.

kito-cheng:
btw, DRIVER_SELF_SPECS can handle option too.

for example:
```
diff --git a/gcc/config/riscv/riscv.h b/gcc/config/riscv/riscv.h
index 4a55ae2..78a848c 100644
--- a/gcc/config/riscv/riscv.h
+++ b/gcc/config/riscv/riscv.h
@@ -131,7 +131,9 @@ along with GCC; see the file COPYING3.  If not see
   {"tune", "%{!mtune=*:-mtune=%(VALUE)}" }, \
   {"float", "%{!msoft-float:%{!mhard-float:-m%(VALUE)-float}}" }, \
 
-#define DRIVER_SELF_SPECS ""
+#define DRIVER_SELF_SPECS \
+  " %{mno-muldiv:-mno-mul -mno-div} " \
+  " %{mmuldiv:-mmul -mdiv} "
 
 #ifdef IN_LIBGCC2
 #undef TARGET_64BIT

```

aswaterman:
My preference would be to avoid DRIVER_SELF_SPECS for option handling.

palmer-dabbelt:
This looks generally good.  @vanjoe: can you address Kito's comments?  We want to keep "-mno-muldiv" and "-mmuldiv", since other stuff has been using them.  You can just force push over this PR and it'll update.

@kito-cheng Thanks for looking at this!

vanjoe:
Updated

gcc/common/config/riscv/riscv-common.c
```diff
@@ -110,7 +110,13 @@ riscv_handle_option (struct gcc_options *opts,
...
       riscv_parse_arch_string (decoded->arg, &opts->x_target_flags);
       return true;
-
+    case OPT_mmuldiv:
+      if (decoded->value){
+      opts->x_target_flags |= (MASK_MUL | MASK_DIV);
+      }else{
+         opts->x_target_flags &= ~(MASK_MUL | MASK_DIV);
+      }
+      return true;
```
palmer-dabbelt: This doesn't match the GNU indenting style.

gcc/config/riscv/riscv.h
```diff
@@ -63,7 +63,11 @@ along with GCC; see the file COPYING3.  If not see
       if (TARGET_ATOMIC)						\
 	builtin_define ("__riscv_atomic");				\
 									\
-      if (TARGET_MULDIV)						\
+      if (TARGET_MUL)						\
+	builtin_define ("__riscv_mul");				\
+      if (TARGET_DIV)						\
+	builtin_define ("__riscv_div");				\
+      if (TARGET_DIV && TARGET_MUL)						\
```
palmer-dabbelt: It's possible the patch has broken these indents, but I think the \'s don't line up.

vanjoe: None of the \'s line up in my emacs. Will fix.

What test suite do you want me to run?

palmer-dabbelt: The newlib test suite should be fine:

https://github.com/riscv/riscv-gnu-toolchain/blob/master/README.md#test-suite

gcc/common/config/riscv/riscv-common.c
```diff
@@ -110,7 +110,16 @@ riscv_handle_option (struct gcc_options *opts,
...
+    case OPT_mmuldiv:
+      if (decoded->value)
+         {
+            opts->x_target_flags |= (MASK_MUL | MASK_DIV);
+         }
+      else
+         {
+            opts->x_target_flags &= ~(MASK_MUL | MASK_DIV);
+         }
+      return true;
```
kito-cheng: brackets can be removed if the `if` or `else` body contain only 1 line.

vanjoe:
Tests report : 
		=== gcc Summary ===

# of expected passes		12790
# of unexpected failures	15
# of expected failures		7
# of unsupported tests		167


Is that normal?

palmer-dabbelt:
It's not what mine looks like, but I probably screwed something up in my own directory.  This is merged

  https://github.com/riscv/riscv-gcc/commit/4aabfe51c4678948f02609d3a898c2a2c2df466f

but github isn't picking it up.

Thanks for the patches!

PR #9, opened 2016-10-18
Empty isa string

kito-cheng:


aswaterman:
approved but will let @palmer-dabbelt merge

palmer-dabbelt:
Whoops, I missed this.  @kito-cheng: is the first commit actually correct?  After looking at it again I think I might have just screwed up?

Either way, I'm closing #10

kito-cheng:
@palmer-dabbelt I've fix `Treat the empty ISA string like G` without comment/author change by git rebase -i in this pull request.

palmer-dabbelt:
Ah, OK, I get it.  I clicked the "rebase and merge" button for the first time, I believe it worked.

Thanks!

PR #10, opened 2016-10-19
By pass hard-float/soft-float to as

palmer-dabbelt:


palmer-dabbelt:
@kito-cheng I think this is all we need to fix the soft/hard float problem?

https://github.com/riscv/riscv-gnu-toolchain/issues/185

palmer-dabbelt:
Sorry, I'm too far behind on me email.

PR #11, opened 2016-10-24
Set long double to 128 bit for RV64

kito-cheng:
Long double is set to 128 bit now!
```
$ cat long-double.c
#include <stdio.h>

long double a, b, c;
int main()
{
  a = 10.2;
  b = 11.3;
  c = a + b;
  printf ("%Lf + %Lf = %Lf\n", a, b, c);
  printf ("sizeof (long double) = %d\n", (int)sizeof(long double));
  return 0;
}
$ riscv64-linux-gcc long-double.c  -static
$ qemu-riscv64 a.out 
10.200000 + 11.300000 = 21.500000
sizeof (long double) = 16
```

ABI:
- Always pass in memory for long double
- Always pass in memory for struct {long double}

Testsutie Result: 
```
                                    |     RV32IMA   |   RV32IMAFD   |     RV64IMA   |   RV64IMAFD   | 
gcc       # of expected failures    |   127 (    0) |   127 (    0) |   129 (    0) |   129 (    0) | 
          # of expected passes      | 82386 (    5) | 82398 (    5) | 83810 (   13) | 83814 (   23) | 
          # of unexpected failures  |    57 (   -5) |    64 (   -5) |    72 (   -5) |   101 (  -19) | 
          # of unexpected successes |     0 (    0) |     0 (    0) |     0 (    0) |     0 (    0) | 
          # of unresolved testcases |    10 (    0) |    10 (    0) |    10 (    0) |    18 (   -2) | 
          # of unsupported tests    |  1645 (    0) |  1636 (    0) |  1516 (    0) |  1499 (    0) | 
g++       # of expected failures    |   303 (    0) |   303 (    0) |   303 (    0) |   303 (    0) | 
          # of expected passes      | 90602 (    0) | 90602 (    0) | 92506 (   24) | 92506 (   24) | 
          # of unexpected failures  |    30 (    0) |    30 (    0) |    22 (    0) |    22 (    0) | 
          # of unexpected successes |     2 (    0) |     2 (    0) |     2 (    0) |     2 (    0) | 
          # of unresolved testcases |     0 (    0) |     0 (    0) |     0 (    0) |     0 (    0) | 
          # of unsupported tests    |  3876 (    0) |  3876 (    0) |  3853 (    0) |  3853 (    0) | 

```


kito-cheng:
glibc is need update too https://github.com/riscv/riscv-glibc/pull/1

kito-cheng:
Add reference https://github.com/riscv/riscv-gnu-toolchain/issues/124

gcc/config/riscv/riscv.c
```diff
@@ -2420,7 +2427,7 @@ riscv_fpr_return_fields (const_tree valtype, tree fields[2])
       if (!SCALAR_FLOAT_TYPE_P (TREE_TYPE (field)))
 	return 0;
 
-      if (i == ARRAY_SIZE (fields))
+      if (i == 2)
```
aswaterman: Can you undo this change?  They are equivalent, but the previous one makes the intent clearer.  (I think this is a merge issue.)

kito-cheng: gcc 6.1 will issue warring for that, and make --enable-werror-always build fail for gcc...

```
riscv.c:2488:31: error: 'sizeof' on array function parameter 'fields' will return size of 'tree_node**' [-Werror=sizeof-array-argument]
```

aswaterman: got it

gcc/config/riscv/riscv.c
```diff
@@ -3609,12 +3630,7 @@ riscv_hard_regno_mode_ok_p (unsigned int regno, enum machine_mode mode)
     }
 
   if (FP_REG_P (regno))
-    {
-      if (mclass == MODE_FLOAT
-	  || mclass == MODE_COMPLEX_FLOAT
-	  || mclass == MODE_VECTOR_FLOAT)
-	return size <= UNITS_PER_HWFPVALUE;
-    }
+    return true;
```
aswaterman: Can you explain this change?

kito-cheng: Floating point register can hold any type of value in theory if target is not force doing normalize by hardware, aarch64 allow floating point register hold any value too[1].

[1] https://github.com/riscv/riscv-gcc/blob/riscv-gcc-6.1.0/gcc/config/aarch64/aarch64.c:859
  - aarch64 has one more checking with hard_regno_nregs, but it's will check automatic by register allocator/IRA, so I think it's unneeded.

kito-cheng: btw, it's also make gpr<-> fpr spilling is possible if register allocator think it's worth (depend on cost model).

aswaterman: That makes sense to me in theory, but I'm concerned about how it interacts with recoding of sub-word floating-point values.  If you spill a 32-bit type using FSD (or FMV.X.D), then the only guarantee is that reloading it with FLD (or FMV.D.X) will re-create the original value.  The value that gets spilled to memory (or an integer register) is implementation-defined.  For example, after executing the code

    fmv.s.x ft0, t0
    fmv.x.d t1, ft0
    fmv.d.x ft1, t1
    fmv.x.s t2, ft1

it is guaranteed that ft0 == ft1, and it is guaranteed that t0 == t2, but it is _not_ guaranteed that t0 == t1.

I think that CANNOT_CHANGE_MODE_CLASS handles this already, but I want to be certain this change won't introduce bugs in certain implementations.

aswaterman:
I'm cool with merging this, once the float mode class thing is reverted, or we ascertain that it is not a problem.  Thanks for the help @kito-cheng

kito-cheng:
I think let float mode class thing revert in this time point until gcc is stable (no regression) since this part like kind of optimization.

And I will update after re-build and re-run testsuite :)

kito-cheng:
Updated, test result no change.

aswaterman:
thanks!!

PR #13, opened 2016-10-26
Revert "Simplify complex float passing code"

kito-cheng:
This reverts commit 1dfe81ea5fdbe089aae6829a06969fe6a7ddc00a.

I propose revert this change since it's make 14 new fail since this commit, it's make complex can't pass half in register and half in stack, that is kind of ABI change, and it's not complete implementation if change ABI is what you want to do.

For example:
```
// Extract from gcc testsuite: gcc/testsuite/gcc.c-torture/execute/complex-7.c
volatile _Complex double d1 = 1.1 + 2.2i;
volatile _Complex double d2 = 3.3 + 4.4i;
volatile _Complex double d3 = 5.5 + 6.6i;
volatile _Complex double d4 = 7.7 + 8.8i;
volatile _Complex double d5 = 9.9 + 10.1i;

// a in a0
// a1 in f1/f2
// a2 in f3/f4
// a3 in f5/f6
// a4 in f7/stack
__attribute__((noinline)) void
check_double (int a, _Complex double a1, _Complex double a2,
              _Complex double a3, _Complex double a4, _Complex double a5)
{
  if (a1 != d1 || a2 != d2 || a3 != d3 || a4 != d4 || a5 != d5)
    abort ();
}
```

Test result after revert:
```
                                    |     RV32IMA   |   RV32IMAFD   |     RV64IMA   |   RV64IMAFD   | 
gcc       # of expected failures    |   127 (    0) |   127 (    0) |   129 (    0) |   129 (    0) | 
          # of expected passes      | 82390 (    0) | 82402 (    0) | 83814 (    0) | 83834 (   16) | 
          # of unexpected failures  |    58 (    0) |    65 (    0) |    73 (    0) |    88 (  -14) | 
          # of unexpected successes |     0 (    0) |     0 (    0) |     0 (    0) |     0 (    0) | 
          # of unresolved testcases |    10 (    0) |    10 (    0) |    10 (    0) |    10 (   -8) | 
          # of unsupported tests    |  1645 (    0) |  1636 (    0) |  1516 (    0) |  1499 (    0) | 
g++       # of expected failures    |   303 (    0) |   303 (    0) |   303 (    0) |   303 (    0) | 
          # of expected passes      | 90602 (    0) | 90602 (    0) | 92506 (    0) | 92506 (    0) | 
          # of unexpected failures  |    30 (    0) |    30 (    0) |    22 (    0) |    22 (    0) | 
          # of unexpected successes |     2 (    0) |     2 (    0) |     2 (    0) |     2 (    0) | 
          # of unresolved testcases |     0 (    0) |     0 (    0) |     0 (    0) |     0 (    0) | 
          # of unsupported tests    |  3876 (    0) |  3876 (    0) |  3853 (    0) |  3853 (    0) | 

```
Note: `expected passes` is +4 compare to https://github.com/riscv/riscv-gcc/pull/11 for each arch since ree optimization is enabled

### DIFF
--- a/gcc/config/riscv/riscv.c
+++ b/gcc/config/riscv/riscv.c
@@ -2341,20 +2341,36 @@ riscv_function_arg (cumulative_args_t cum_v, enum machine_mode mode,
 	}
     }
 
-  /* Pass complex floating-point arguments in FPR pairs, with the real part
-     in the lower register and the imaginary part in the upper register.  */
-  if (info.fpr_p && GET_MODE_CLASS (mode) == MODE_COMPLEX_FLOAT)
+  /* Handle the n32/n64 conventions for passing complex floating-point
+     arguments in FPR pairs.  The real part goes in the lower register
+     and the imaginary part goes in the upper register.  */
+  if (info.fpr_p
+      && GET_MODE_CLASS (mode) == MODE_COMPLEX_FLOAT)
     {
       rtx real, imag;
-      enum machine_mode inner = GET_MODE_INNER (mode);
-      unsigned int regno = FP_ARG_FIRST + info.reg_offset;
+      enum machine_mode inner;
+      unsigned int regno;
 
-      gcc_assert (info.stack_words == 0 && info.reg_words == 2);
-      real = gen_rtx_EXPR_LIST (VOIDmode, gen_rtx_REG (inner, regno),
-				const0_rtx);
-      imag = gen_rtx_EXPR_LIST (VOIDmode, gen_rtx_REG (inner, regno + 1),
-				GEN_INT (GET_MODE_SIZE (inner)));
-      return gen_rtx_PARALLEL (mode, gen_rtvec (2, real, imag));
+      inner = GET_MODE_INNER (mode);
+      regno = FP_ARG_FIRST + info.reg_offset;
+      if (info.reg_words * UNITS_PER_WORD == GET_MODE_SIZE (inner))
+	{
+	  /* Real part in registers, imaginary part on stack.  */
+	  gcc_assert (info.stack_words == info.reg_words);
+	  return gen_rtx_REG (inner, regno);
+	}
+      else
+	{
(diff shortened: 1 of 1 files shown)


aswaterman:
I'll look into this now.  I believe the old code is also subtly incorrect (I think it assumes UNITS_PER_WORD >= UNITS_PER_FP_REG, which is not true for RV32G).

Re: +4 expected passes after REE optimization: this is concerning.  Can you find one of the test cases that previously failed, and try to figure out why it started succeeding?

kito-cheng:
+4 expected passes after enable REE optimization is because there is a test case in gcc testsuite is scanning all dump files which generate by `-fdump-rtl-all`, so if you add 1 more optimization pass, # of expected passes will increase too.

GCC testsuite fragment:
```
...
PASS: gcc.c-torture/unsorted/dump-noaddr.c.*r.postreload,  -O2  comparison
PASS: gcc.c-torture/unsorted/dump-noaddr.c.*r.split2,  -O2  comparison
PASS: gcc.c-torture/unsorted/dump-noaddr.c.*r.ree,  -O2  comparison
PASS: gcc.c-torture/unsorted/dump-noaddr.c.*r.pro_and_epilogue,  -O2  comparison
PASS: gcc.c-torture/unsorted/dump-noaddr.c.*r.dse2,  -O2  comparison
...
```

aswaterman:
OK, cool, thanks for looking into it.

aswaterman:
Yes, both solutions are buggy.  ba211d000e594dd465cef0551cc50fa141b1bea2 reverts the commit and hopefully fixes the bugs in the old one (UNITS_PER_WORD vs. UNITS_PER_FP_REG, and a MIPS assumption about passing long doubles in FPR pairs).  Can you re-test?

kito-cheng:
@aswaterman test result same as this PR :)

aswaterman:
Great!

PR #34, opened 2016-12-05
Change back TARGET_HAVE_TLS to HAVE_AS_TLS since elf toolchain no tls support

kito-cheng:
We need TARGET_HAVE_TLS depend on HAVE_AS_TLS for elf toolchain, otherwise linker will complain __tls_get_addr not found.

### DIFF
--- a/gcc/config/riscv/riscv.c
+++ b/gcc/config/riscv/riscv.c
@@ -4096,8 +4096,10 @@ riscv_cannot_copy_insn_p (rtx_insn *insn)
 #undef TARGET_SCALAR_MODE_SUPPORTED_P
 #define TARGET_SCALAR_MODE_SUPPORTED_P riscv_scalar_mode_supported_p
 
+#ifdef HAVE_AS_TLS
 #undef TARGET_HAVE_TLS
 #define TARGET_HAVE_TLS true
+#endif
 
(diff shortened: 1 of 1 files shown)


aswaterman:
These days, the majority of architectures (including x86 and AArch64) define it to "true".  I think maybe the problem is somewhere else?  Why is __tls_get_addr not defined?

kito-cheng:
Humm, you are right, let me investigate again what's wrong.

kito-cheng:
Code fragment in i386.c
```
#ifdef HAVE_AS_TLS
#undef TARGET_HAVE_TLS
#define TARGET_HAVE_TLS true
#endif
```

Code fragment in aarch64.c
```
#ifdef HAVE_AS_TLS
#undef TARGET_HAVE_TLS
#define TARGET_HAVE_TLS 1
#endif
```

Code fragment in arm.c
```
#ifdef HAVE_AS_TLS
#undef TARGET_HAVE_TLS
#define TARGET_HAVE_TLS true
#endif
```

aswaterman:
Ah, I see.  We also needed to fix the configure.ac, which had an out-of-date test program to determine HAVE_AS_TLS.  It's done, thanks for reporting.

PR #37, opened 2016-12-05
Refined predefined macro proposal [do not merge]

aswaterman:
This is a placeholder to get feedback.

### DIFF
--- a/gcc/config/riscv/riscv.h
+++ b/gcc/config/riscv/riscv.h
@@ -28,19 +28,7 @@ along with GCC; see the file COPYING3.  If not see
 #define TARGET_CPU_CPP_BUILTINS()					\
   do									\
     {									\
-      builtin_assert ("machine=riscv");					\
-      builtin_assert ("cpu=riscv");					\
-      builtin_define ("__riscv__");					\
       builtin_define ("__riscv");					\
-									\
-      if (TARGET_64BIT)							\
-	builtin_define ("__riscv64");					\
-      else								\
-	builtin_define ("__riscv32");					\
-									\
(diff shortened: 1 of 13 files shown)


aswaterman:
@sorear

sorear:
1. Why are we adding trailing underscores, I thought the agreement was to get rid of them everywhere?
2. Does `__riscv_xlen__ == 64` become the way to distinguish RV32 from RV64ILP32?
3. There are few very bad predefined macros which might be coming from somewhere else, like `_riscv` (redundant *and* not within the ANSI C reserved namespace)

sorear:
RV32: `__riscv_xlen == 32`
RV64: `__riscv_xlen == 64 && __SIZEOF_POINTER__ == 8`
RV64 ILP32: `__riscv_xlen == 64 && __SIZEOF_POINTER__ == 4`

Viable?

(I was originally thinking we'd use `__riscv64` for this, but `__riscv_xlen` works as well.)

aswaterman:
Makes sense to me.

aswaterman:
Updated the PR accordingly.

aswaterman:
This is the RV64G compiler's output now:

    #define __riscv 1
    #define __riscv_atomic 1
    #define __riscv_cmodel_medlow 1
    #define __riscv_fdiv 1
    #define __riscv_float_abi_double 1
    #define __riscv_flen 64
    #define __riscv_mul 1
    #define __riscv_muldiv 1
    #define __riscv_xlen 64
    #define __riscv_fsqrt 1
    #define __riscv_div 1

palmer-dabbelt:
LGTM

PR #39, opened 2016-12-14
Don't build the F multilibs on Linux

palmer-dabbelt:
We've decided that we're not going to add F to the default multilib set on Linux, since it's unlikely that general purpose hardware will have single-precision but won't have double precision.  We'll still support the non-multilib {lp64,ilp32}f ABIs on all -march targets, and distributions are free to build them if they think they're interesting for their use cases, they just won't be in the default Linux make fragment.

@kito-cheng This also reverts your recent commit that I merged -- since we won't have an F multilib then users shouldn't get it by default.

### DIFF
--- a/gcc/config.gcc
+++ b/gcc/config.gcc
@@ -4019,10 +4019,8 @@ case "${target}" in
 		"")
 			case "${with_arch}" in
 			rv32*d* | rv32g*) with_abi=ilp32d ;;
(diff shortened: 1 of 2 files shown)


aswaterman:
This makes sense to me.  I had always imagined the soft-float ABI being the default for RVF, with the option of an F ABI for specialized systems.

PR #40, opened 2016-12-16
Fix riscv_flatten_aggregate_field for handle array of empty struct

kito-cheng:
- Test case:
```
struct S12 { struct{}a[0]; } ;
extern struct S12 s12;
struct S12 a12[5];
struct S12 check12 (struct S12 arg0, struct S12 *arg1, struct S12 arg2)
{
  struct S12 ret;
  return ret;
}
```

### DIFF
--- a/gcc/config/riscv/riscv.c
+++ b/gcc/config/riscv/riscv.c
@@ -2217,7 +2217,7 @@ riscv_flatten_aggregate_field (const_tree type,
 							 subfields, 0, offset);
 
 	/* Can't handle incomplete types nor sizes that are not fixed.  */
-	if (n_subfields < 0
+	if (n_subfields <= 0
 	    || !COMPLETE_TYPE_P (type)
 	    || TREE_CODE (TYPE_SIZE (type)) != INTEGER_CST
 	    || !index


aswaterman:
Thanks for identifying the corner case!  What happened previously, and what is the new behavior?

kito-cheng:
It's will hang for long time, make gcc testsuite timeout, new behavior is just return -1 since empty struct not pass in register (and not pass in memory too).

n_elts = 4294967296 and n_subfields = 0, so nested loop never execute but it run outer loop 4294967296 times.
```
  2235         for (HOST_WIDE_INT i = 0; i < n_elts; i++)
  2236           for (int j = 0; j < n_subfields; j++)
  2237             {
  2238               if (n >= 2)
  2239                 return -1;
  2240 
  2241               fields[n] = subfields[j];
  2242               fields[n++].offset += i * tree_to_uhwi (elt_size);
  2243             }
```

aswaterman:
Sorry, my question was unclear.  After this change, how does the ABI pass this struct?

kito-cheng:
Hummmmm, before and after is pass in nowhere too, I think we can add a rule for empty struct into riscv-elf-psabi-doc.

aswaterman:
Pass-nowhere makes sense to me.  @sorear should comment on the ABI issue.

Since this patch is a bugfix, not an ABI change, I'm going to merge it anyway.  Thanks, Kito.

sorear:
Empty structs was something I was thinking about when we were discussing the calling convention and I thought we were doing pass-nowhere at that time, e.g. (int x, struct { } y, int z) passes in a0 and a1.  If that's what we're doing now, then I have no problem.

(You probably know this already but the behavior is different in C++ which requires a minimum sizeof of 1 for all types)

PR #43, opened 2017-01-06
Split the RISC-V target makefrags into multilib and nomultilib

palmer-dabbelt:
This is the best way I can figure out how to make this work: we now have
two target makefile fragments for each RISC-V OS, one that for multilib
support and one for non-multilib.  The only difference is that the
multilib one sets MULTILIB_REQUIRED, as setting that seems to screw up
the non-multilib paths.

The other option would be to modify something like LINK_SPEC, but I
think this is better because it allows us to later add to the multilibs
that we're building and transparently support those new ones.

### DIFF
--- a/gcc/config.gcc
+++ b/gcc/config.gcc
@@ -2035,7 +2035,11 @@ microblaze*-*-elf)
         ;;
 riscv*-*-linux*)
 	tm_file="elfos.h gnu-user.h linux.h glibc-stdint.h ${tm_file} riscv/linux.h"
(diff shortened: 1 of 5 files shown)


aswaterman:
I agree that screwing with LINK_SPEC seems worse.

I'll look at this tomorrow; too tired now.

palmer-dabbelt:
@aswaterman Thoughts?

aswaterman:
Looks OK to me.  If there is a better solution, it might come up in upstreaming.

PR #45, opened 2017-01-22
Add nfma<mode>4_unsafe and nfms<mode>4_unsafe pattern

kito-cheng:
This patch was hold in my branch for few days, I just not sure does it need guard by fast-math or not, however I've read https://groups.google.com/a/groups.riscv.org/forum/#!msg/sw-dev/795mx1ZQNFY/-LmhAJzIFwAJ this thread, so... @aswaterman do you think we just match this pattern without -ffast-math?
and this patch can fix the issue https://github.com/riscv/riscv-gnu-toolchain/issues/211 .

### DIFF
--- a/gcc/config/riscv/riscv.md
+++ b/gcc/config/riscv/riscv.md
@@ -846,6 +846,30 @@
   [(set_attr "type" "fmadd")
    (set_attr "mode" "<UNITMODE>")])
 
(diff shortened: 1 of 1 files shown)


aswaterman:
I think we should match these without -ffast-math; can you update the PR?

aswaterman:
I'll pull this into submission branch manually.

PR #46, opened 2017-01-22
Fix for review's comment

kito-cheng:


aswaterman:
We need to stay in sync better--I already did most of this tonight. I'll review and merge our changes tomorrow.

aswaterman:
We took the same approach, so I'll manually merge and push.

Going forward, can you let me know what you're working on, so we can avoid duplication of effort?

PR #48, opened 2017-01-31
Fix for review's comment

kito-cheng:


libgcc/config/riscv/save-restore.S
```diff
@@ -60,55 +60,107 @@ __riscv_save_12:
...
   sd s2, 80(sp)
+  .cfi_offset 18, -32
   sd s1, 88(sp)
+  .cfi_offset 9, -24
   sd s0, 96(sp)
+  .cfi_offset 8, -16
   sd ra, 104(sp)
+  .cfi_offset 1, -8
+  # CFA info is not correct in next 2 instruction since t1's
+  # value is depend on how may register really save.
```
aswaterman: Does that mean that CFA info is also incorrect once we return to the caller?

kito-cheng: No, the caller's CFA is maintained by gcc, so it's only got wrong back-trace in next 2 instruction, and get correct again once it's return.

aswaterman: got it.

sorear: If I'm understanding right this will confuse statistical profilers so it'd be somewhat useful to have working unwind here.  (I do not regard this as upstream-blocking to be clear)  Would fixing this require an arch-specific unwind opcode?

kito-cheng: Hmm, let me discuss with our binutils guy tomorrow.

PR #55, opened 2017-03-13
Use more conservative fences on RISC-V

palmer-dabbelt:
The RISC-V memory model is still in the process of being formally
specified, so for now we're going to be safe and add the I/O bits to
userspace fences because there's no way to know if userspace is touching
memory-mapped I/O regions at compile time.

This will have no impact on existing microarchitecutres because they
treat all fences conservatively.

gcc/ChangeLog:

        * config/riscv/riscv.c (riscv_print_operand): Use "fence
        rwio,wo".
        * config/riscv/sync.mc (mem_thread_fence_1): Use "fence
        rwio,rwio".

### DIFF
--- a/gcc/ChangeLog
+++ b/gcc/ChangeLog
@@ -1,3 +1,10 @@
+2017-03-13  Palmer Dabbelt  <palmer@dabbelt.com
+
+	* config/riscv/riscv.c (riscv_print_operand): Use "fence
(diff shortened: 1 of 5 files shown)


palmer-dabbelt:
I think this is the only way to fix https://github.com/riscv/riscv-gcc/issues/54.

PR #56, opened 2017-03-29
Clean up and fix indention for riscv.md

kito-cheng:


palmer-dabbelt:
Thanks Kito.  I'm assuming the first patch doesn't change any functionality (and you've verified this by making sure it still passes the test suite)?  It's big and I don't want to go through it line-by-line.

kito-cheng:
Yes, it's no functional change

kito-cheng:
I've run rv32ima and it's ok

PR #59, opened 2017-04-01
RISC-V: Add -mstrict-align option

aswaterman:
2017-04-30  Andrew Waterman  <andrew@sifive.com>

	* config/riscv/riscv.opt (mstrict-align): New option.
	* config/riscv/riscv.h (STRICT_ALIGNMENT): Use it.  Update comment.
	(SLOW_UNALIGNED_ACCESS): Define.
	(riscv_slow_unaligned_access): Declare.
	* config/riscv/riscv.c (riscv_tune_info): Add slow_unaligned_access
	field.
	(riscv_slow_unaligned_access): New variable.
	(rocket_tune_info): Set slow_unaligned_access to true.
	(optimize_size_tune_info): Set slow_unaligned_access to false.
	(riscv_cpu_info_table): Add entry for optimize_size_tune_info.
	(riscv_valid_lo_sum_p): Use TARGET_STRICT_ALIGN.
	(riscv_option_override): Set riscv_slow_unaligned_access.

### DIFF
--- a/gcc/config/riscv/riscv.c
+++ b/gcc/config/riscv/riscv.c
@@ -255,6 +255,7 @@ struct riscv_tune_info
   unsigned short issue_rate;
   unsigned short branch_cost;
   unsigned short memory_cost;
+  bool slow_unaligned_access;
 };
 
 /* Information about one CPU we know about.  */
@@ -267,6 +268,9 @@ struct riscv_cpu_info {
 };
 
 /* Global variables for machine-dependent things.  */
+
+/* Whether unaligned accesses execute very slowly.  */
+bool riscv_slow_unaligned_access;
 
 /* Which tuning parameters to use.  */
 static const struct riscv_tune_info *tune_info;
@@ -301,7 +305,8 @@ static const struct riscv_tune_info rocket_tune_info = {
   {COSTS_N_INSNS (6), COSTS_N_INSNS (6)},	/* int_div */
   1,						/* issue_rate */
   3,						/* branch_cost */
-  5						/* memory_cost */
+  5,						/* memory_cost */
+  true,						/* slow_unaligned_access */
 };
 
 /* Costs to use when optimizing for size.  */
@@ -313,12 +318,14 @@ static const struct riscv_tune_info optimize_size_tune_info = {
   {COSTS_N_INSNS (1), COSTS_N_INSNS (1)},	/* int_div */
   1,						/* issue_rate */
   1,						/* branch_cost */
-  2						/* memory_cost */
+  2,						/* memory_cost */
+  false,					/* slow_unaligned_access */
 };
 
 /* A table describing all the processors GCC knows about.  */
 static const struct riscv_cpu_info riscv_cpu_info_table[] = {
   { "rocket", &rocket_tune_info },
+  { "size", &optimize_size_tune_info },
 };
 
 /* Return the riscv_cpu_info entry for the given name string.  */
@@ -726,7 +733,8 @@ riscv_valid_lo_sum_p (enum riscv_symbol_type sym_type, enum machine_mode mode)
   /* We may need to split multiword moves, so make sure that each word
      can be accessed without inducing a carry.  */
   if (GET_MODE_SIZE (mode) > UNITS_PER_WORD
-      && GET_MODE_BITSIZE (mode) > GET_MODE_ALIGNMENT (mode))
+      && (!TARGET_STRICT_ALIGN
+	  || GET_MODE_BITSIZE (mode) > GET_MODE_ALIGNMENT (mode)))
     return false;
 
   return true;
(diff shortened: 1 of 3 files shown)


aswaterman:
@sorear @kito-cheng @palmer-dabbelt when optimizing for size (-Os), this PR allows misaligned accesses by default, because they _do_ reduce code size.  Do you think that's the right call?

sorear:
My gut reaction is:

* `STRICT_ALIGNMENT` should be 0 by default, but changed by `-mstrict-alignment` for people who are working in M-mode and don't want to set up a trap handler
* `SLOW_UNALIGNED_ACCESS` should depend on the tune setting
* Rocket's misalignment traps are too slow for `SLOW_UNALIGNED_ACCESS=0` even at `-Os`

But I have no real idea of what the implications of that would be (both for functionality and compatibility), and I am +0 on this PR.

aswaterman:
Hmm... given the existence of ```SLOW_UNALIGNED_ACCESS```, it does feel a little wrong to have ```-mtune``` affect a correctness setting like ```STRICT_ALIGNMENT```.

aswaterman:
PR updated accordingly.  Running test suite.

aswaterman:
I think the struct-by-value-8 fails as a result of this, but I don't have time to look into it for the next few days.  If someone else gets around to it first, let me know.

palmer-dabbelt:
I think there should be two settings here:

 * A tuning flag that says "is doing a misaligned access fast?"
 * A correctness flag that says "is doing a misaligned access legal?"

While it seems sane to have "-Os" default to "emit unaligned accesses for performance reasons", it would also be good to have a way to turn that off.

PowerPC calls the "is doing a misaligned access legal" flag "-mstrict-align", so I think copying that is the right thing to do.  We could call the other one something like "-mtune-align"?

aswaterman:
Agreed re: -mstrict-align.

Re: -mtune-align, why not just rely on -mtune=CPU? In other words, which
tuning parameters are important enough to get their own flags?

palmer-dabbelt:
I thought (at least on x86), that every tunable could be set independently from a command-line flag and that "-mtune" just provided a set of those.  Looking at

````
gcc -march=native -Q --help=target
````

it looks like I'm wrong about that, so I guess there's no reason to have more stuff in our port.

aswaterman:
@sorear I guess we should conditionally define __riscv_strict_align?

sorear:
@aswaterman Probably a good idea, although I'm on the fence about whether we're sending too many signals that this is supported

aswaterman:
The struct-by-value-8 test fails in RV32IMA[FD], but only when dynamically linked.  When statically linked (identical assembly), it works fine.  That makes me suspect a qemu bug, but since it doesn't happen for RV64, I can't try it in Linux on Spike.

sorear:
@aswaterman this is getting old.  possible to send me a set of binaries, or point me at the source or something so I can take a poke at qemu?

aswaterman:
Thanks for staying on this.  Here are four binaries (static & dynamic linking, strict and not-strict).  All but dynamic, not-strict pass.

[struct-by-value-8-rv32ima.zip](https://github.com/riscv/riscv-gcc/files/966434/struct-by-value-8-rv32ima.zip)

sorear:
@aswaterman So I just built a fresh riscv-gnu-toolchain with --enable-multilib and a fresh qemu-riscv32 and I'm getting an immediate SIGILL for all of the binaries.  Any idea what's going on?

aswaterman:
No idea, I'll bump my qemu to latest to see if there was a regression there.

sorear:
@aswaterman Found it, I broke something in riscv/riscv-qemu#59 and qemu was starting with all extensions disabled (SIGILL on first compressed instruction).  Trying qemu master branch now

aswaterman:
@sorear identified the symptom: the compiler assumed that sub-alignment offsets into aligned objects would not overflow past the alignment granule.  The MIPS compiler makes the same assumption, which suggests it's fundamentally correct for STRICT_ALIGNMENT targets.

I've updated the PR so that riscv_valid_lo_sum_p only makes the assumption when -mstrict-align is passed.  Re-running regression tests now.

aswaterman:
@palmer-dabbelt this is now back to status quo on the test suite, and I've updated the PR description to match my suggested patch description.

palmer-dabbelt:
https://gcc.gnu.org/ml/gcc-patches/2017-05/msg00011.html

@aswaterman FYI, this needs a doc change

````
diff --git a/gcc/doc/invoke.texi b/gcc/doc/invoke.texi
index 0eeea7b..9ae469a 100644
--- a/gcc/doc/invoke.texi
+++ b/gcc/doc/invoke.texi
@@ -975,6 +975,7 @@ See RS/6000 and PowerPC Options.
 -mtune=@var{processor-string} @gol
 -msmall-data-limit=@var{N-bytes} @gol
 -msave-restore  -mno-save-restore @gol
+-mstrict-align -mno-strict-align @gol
 -mcmodel=@var{code-model} @gol
 -mexplicit-relocs  -mno-explicit-relocs @gol}
 
@@ -20945,6 +20946,11 @@ Put global and static data smaller than @var{n} bytes into a special section
 @opindex msave-restore
 Use smaller but slower prologue and epilogue code.
 
+@item -mstrict-align
+@itemx -mno-strict-align
+@opindex mstrict-align
+Assume that unaligned memory accesses are disallowed.
+
 @item -mcmodel=@var{code-model}
 @opindex mcmodel
 Specify the code model.
````

PR #60, opened 2017-04-03
Add optimised 64-bit byteswap for riscv64

michaeljclark:
This implementation is 20 instructions and accesses two constants versus 30 instructions for the C version. Might be an idea to benchmark on Rocket before merging.

### DIFF
--- a/gcc/config/riscv/riscv.md
+++ b/gcc/config/riscv/riscv.md
@@ -398,47 +398,47 @@
 ;;
 
 (define_insn "add<mode>3"
-  [(set (match_operand:ANYF 0 "register_operand" "=f")
-	(plus:ANYF (match_operand:ANYF 1 "register_operand" "f")
-		   (match_operand:ANYF 2 "register_operand" "f")))]
+  [(set (match_operand:ANYF            0 "register_operand" "=f")
+	(plus:ANYF (match_operand:ANYF 1 "register_operand" " f")
+		   (match_operand:ANYF 2 "register_operand" " f")))]
   "TARGET_HARD_FLOAT"
   "fadd.<fmt>\t%0,%1,%2"
   [(set_attr "type" "fadd")
    (set_attr "mode" "<UNITMODE>")])
 
 (define_insn "addsi3"
-  [(set (match_operand:SI 0 "register_operand" "=r,r")
-	(plus:SI (match_operand:SI 1 "register_operand" "r,r")
-		  (match_operand:SI 2 "arith_operand" "r,I")))]
+  [(set (match_operand:SI          0 "register_operand" "=r,r")
+	(plus:SI (match_operand:SI 1 "register_operand" " r,r")
+		 (match_operand:SI 2 "arith_operand"    " r,I")))]
   ""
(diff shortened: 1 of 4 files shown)


michaeljclark:
I added the symbol `_bswapdi2` to `libgcc/config/riscv/t-elf64` and have `#if __riscv_xlen == 64` in `libgcc/config/riscv/bswapdi2.S` and have tested a clean build of the compiler on rv64. Still need to test that it doesn't interfere with the rv32 build.

This is output from an executable built using `riscv64-unknown-elf-gcc` with this patch

```
000000000001080c <__bswapdi2>:
   1080c:	0000d717          	auipc	a4,0xd
   10810:	bad73703          	ld	a4,-1107(a4) # 1d3b9 <__bswapdi2_c1>
   10814:	00851793          	slli	a5,a0,0x8
   10818:	8ff9                	and	a5,a5,a4
   1081a:	8121                	srli	a0,a0,0x8
   1081c:	8321                	srli	a4,a4,0x8
   1081e:	8d79                	and	a0,a0,a4
   10820:	8fc9                	or	a5,a5,a0
   10822:	0000d717          	auipc	a4,0xd
   10826:	b9f73703          	ld	a4,-1121(a4) # 1d3c1 <__bswapdi2_c2>
   1082a:	01079513          	slli	a0,a5,0x10
   1082e:	8d79                	and	a0,a0,a4
   10830:	83c1                	srli	a5,a5,0x10
   10832:	8341                	srli	a4,a4,0x10
   10834:	8ff9                	and	a5,a5,a4
   10836:	8fc9                	or	a5,a5,a0
   10838:	02079513          	slli	a0,a5,0x20
   1083c:	9381                	srli	a5,a5,0x20
   1083e:	8d5d                	or	a0,a0,a5
   10840:	8082                	ret
```

aswaterman:
I think we don't need the __riscv_xlen check, and I think we need to do something different than .rodata to make the loads relax to gp.  I believe that strcmp.S in glibc has an example of how to do that.

michaeljclark:
OK. I think the patch needs a bit of testing first. I'll have a look at strcmp.S ...

michaeljclark:
I think this should keep this on hold. I'm still unsure whether we should introduce a constant read from dcache (i.e. require LOAD) in an intrinsic like bswap. The current version doesn't touch dcache or read from constant data (ROM).  It may be slower on some architectures depending on where the constant data is stored e.g. HiFive1. It's ultimately going to be addressed with the B extension...

I still need to look at strcmp.S regarding the GP relative relaxation, if we decide to proceed with a change like this...

aswaterman:
For RV64, GCC already freely emits loads to the constant pool for constants that take many instructions to synthesize in-line.  So there's a built-in expectation that they are at least reasonably performant.  (On HiFive1, for example, they can be allocated to the data scratchpad instead of off-chip flash.)  So, I don't think it's making an unsafe assumption for RV64 library code to do the same.

But feel free to close the issue if you think it's best to table it for now.

michaeljclark:
Closing this for the moment.

PR #62, opened 2017-04-05
[WIP] Improve(?) cost model

aswaterman:
Let's not merge this yet, since we should run it on more than Dhrystone :)


palmer-dabbelt:
@aswaterman I think this is defunct?

aswaterman:
Maybe, but can you first verify it doesn't help anymore?

PR #66, opened 2017-05-03
Work around for commit_one_edge_insertion's assertion fail

kito-cheng:
@aswaterman I guess you will need this work around.

### DIFF
--- a/gcc/config/riscv/riscv.c
+++ b/gcc/config/riscv/riscv.c
@@ -2786,6 +2786,11 @@ riscv_block_move_loop (rtx dest, rtx src, HOST_WIDE_INT length,
   /* Mop up any left-over bytes.  */
   if (leftover)
     riscv_block_move_straight (dest, src, leftover);
+  else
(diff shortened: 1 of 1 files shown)


kito-cheng:
Relate issue for this change: https://github.com/riscv/riscv-gcc/issues/21

aswaterman:
Maybe we should just kill riscv_block_move_loop and only do movmemsi when it's straight-line code.  Once you add the loop overhead, the cost of doing memcpy isn't so bad.

kito-cheng:
movmemsi not only gain from the straight-line version but also gain from `inline` memcpy, so loop version is still worth to gen even it's same as memcpy function in my experiences.

kito-cheng:
it's can save cost for function call: a jal and cost for caller save registers.

aswaterman:
Yeah, you're right, I'll leave it alone.

aswaterman:
The ```nop``` instruction will be killed by dead-code elimination, right?

kito-cheng:
>The nop instruction will be killed by dead-code elimination, right?
Yes :)

PR #71, opened 2017-05-18
Linux isn't booting

palmer-dabbelt:
@kito-cheng it looks like your latest patch broken the Linux boot.  I'm not going to have time to look at it for a bit, do you mind taking a look for me?

This PR has some commits to work around the bug, but I'd still like a better fix than the T6_REGNUM one, if possible.

### DIFF
--- a/gcc/config/riscv/pic.md
+++ b/gcc/config/riscv/pic.md
@@ -39,17 +39,17 @@
 (define_insn "*local_pic_load<mode>"
   [(set (match_operand:ANYF 0 "register_operand" "=f")
 	(mem:ANYF (match_operand 1 "absolute_symbolic_operand" "")))
-   (clobber (match_scratch:DI 2 "=r"))]
+   (clobber (reg:DI T6_REGNUM))]
   "TARGET_HARD_FLOAT && TARGET_64BIT && USE_LOAD_ADDRESS_MACRO (operands[1])"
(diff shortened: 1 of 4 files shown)


gcc/config/riscv/pic.md
```diff
@@ -39,17 +39,17 @@
...
   "TARGET_HARD_FLOAT && TARGET_64BIT && USE_LOAD_ADDRESS_MACRO (operands[1])"
-  "<load>\t%0,%1,%2"
+  "<load>\t%0,%1,t6"
   [(set (attr "length") (const_int 8))])
 
 (define_insn "*local_pic_load<mode>"
   [(set (match_operand:ANYF 0 "register_operand" "=f")
 	(mem:ANYF (match_operand 1 "absolute_symbolic_operand" "")))
-   (clobber (match_scratch:SI 2 "=r"))]
+   (clobber (reg:DI T6_REGNUM))]
```
aswaterman: This might be an old bug, but the mode should be ```SI```, not ```DI```, since the pattern is for ```!TARGET_64BIT```.

palmer-dabbelt: Thanks, it's probably just a copy/paste error on my end.  I cherry-pick'd this patch without looking at it, and I haven't tested it on RV32 yet.  I'll fix it up.

PR #73, opened 2017-05-18
Don't use slow unaligned accesses in store-merging pass

aswaterman:
2017-05-18  Andrew Waterman  <andrew@sifive.com>

	* config/riscv/riscv.c: Include params.h.
	(riscv_option_override): Set PARAM_STORE_MERGING_ALLOW_UNALIGNED.

### DIFF
--- a/gcc/ChangeLog
+++ b/gcc/ChangeLog
@@ -1,3 +1,8 @@
+2017-05-18  Andrew Waterman  <andrew@sifive.com>
+
+	* config/riscv/riscv.c (riscv_option_override): Conditionally set
+	TARGET_STRICT_ALIGN based upon -mtune argument.
+
 2017-05-12  Kito Cheng  <kito.cheng@gmail.com>
 
 	* config/riscv/riscv.c (riscv_legitimize_move): Handle
--- a/gcc/config/riscv/riscv.c
+++ b/gcc/config/riscv/riscv.c
@@ -3799,9 +3799,13 @@ riscv_option_override (void)
 
   /* Use -mtune's setting for slow_unaligned_access, even when optimizing
      for size.  For architectures that trap and emulate unaligned accesses,
-     the performance cost is too great, even for -Os.  */
+     the performance cost is too great, even for -Os.  Similarly, if
+     -m[no-]strict-align is left unspecified, heed -mtune's advice.  */
   riscv_slow_unaligned_access = (cpu->tune_info->slow_unaligned_access
 				 || TARGET_STRICT_ALIGN);
+  if ((target_flags_explicit & MASK_STRICT_ALIGN) == 0
+      && cpu->tune_info->slow_unaligned_access)
+    target_flags |= MASK_STRICT_ALIGN;
 
   /* If the user hasn't specified a branch cost, use the processor's
      default.  */


aswaterman:
This is arguably a workaround for a bug in the store-merging pass (I think they should be checking SLOW_UNALIGNED_ACCESS instead of checking STRICT_ALIGN), but this patch is obviously less controversial.

palmer-dabbelt:
LGTM.  I'll run some tests.

kito-cheng:
We've hit this issue on this week too...:(

kito-cheng:
I just dig out where use `STRICT_ALIGNMENT` and `SLOW_UNALIGNED_ACCESS` in GCC, and seem some part only check `STRICT_ALIGNMENT` and may generate unaligned access or if `!STRICT_ALIGNMENT`, so I think maybe make SLOW_UNALIGNED_ACCESS and STRICT_ALIGNMENT couple again is better way?

aswaterman:
OK, so your proposal is:

- If ```-mstrict-align```, set STRICT_ALIGN=SLOW_UNALIGNED_ACCESS=1.
- If ```-mno-strict-align```, set STRICT_ALIGN=SLOW_UNALIGNED_ACCESS=(whatever ```-mtune``` says).

I think this would work, but it seems like these are bugs in mainline GCC... maybe we should just fix the mainline bugs?

kito-cheng:
Humm, 
- If -mstrict-align, set STRICT_ALIGN=SLOW_UNALIGNED_ACCESS=1.
- If -mno-strict-align, set STRICT_ALIGN=SLOW_UNALIGNED_ACCESS=0
- If -mstrict-align and -mno-strict-align not given,  set STRICT_ALIGN=SLOW_UNALIGNED_ACCESS=(whatever -mtune says).

I think GCC's logic is if STRICT_ALIGN is 0 then mean it's OK for generate unaligned access, and SLOW_UNALIGNED_ACCESS can't used during middle-end (GIMPLE) since GCC don't know the accurate machine mode (and SLOW_UNALIGNED_ACCESS take an argument for machine mode), so set STRICT_ALIGN to 0 and SLOW_UNALIGNED_ACCESS always return 1 seem is conflict for GCC I guess.

[1] https://gcc.gnu.org/onlinedocs/gccint/Costs.html#index-SLOW_005fUNALIGNED_005fACCESS

aswaterman:
Yeah, last night, I came to the same conclusion about -mstrict-align and -mno-strict-align not given.

Fair point about the machine mode.  I will make this change.

aswaterman:
Actually, slightly different proposal:

- If -mstrict-align, then STRICT_ALIGN=SLOW_UNALIGNED_ACCESS=1
- If -mno-strict-align, then STRICT_ALIGN=0 and SLOW_UNALIGNED_ACCESS=-mtune
- If neither, then STRICT_ALIGN=SLOW_UNALIGNED_ACCESS=-mtune

This makes it possible to exhibit all behaviors, which I think is what we actually want.

aswaterman:
OK, I've updated the PR to reflect this proposal.  If we are _really_ sure that STRICT_ALIGN should always equal SLOW_UNALIGNED_ACCESS, then we should go further and remove ```riscv_slow_unaligned_access``` and then ```#define SLOW_UNALIGNED_ACCESS(MODE, ALIGN) STRICT_ALIGN```.

I see that rs6000 has STRICT_ALIGN=0 and sometimes SLOW_UNALIGNED_ACCESS != 0, but as you said, it appears to be used for tuning block moves.

kito-cheng:
LGTM and no regression for elf toolchain with rv32imac, rv32gc, rv64imac and rv64gc.

PR #77, opened 2017-06-16
Support RV32E

kito-cheng:


gcc/config/riscv/riscv.h
```diff
@@ -513,15 +513,15 @@ enum reg_class
    `crtl->outgoing_args_size'.  */
 #define OUTGOING_REG_PARM_STACK_SPACE(FNTYPE) 1
 
-#define STACK_BOUNDARY 128
+#define STACK_BOUNDARY (TARGET_RVE ? 32 : 128)
```
palmer-dabbelt: @aswaterman Won't we still need the stack to be 16-byte aligned because of the C.ADDI16SP instruction?  This would change BIGGEST_ALIGNMENT as well.

gcc/config/riscv/riscv.h
```diff
@@ -513,15 +513,15 @@ enum reg_class
...
 

 /* Symbolic macros for the registers used to return integer and floating
    point values.  */
 
 #define GP_RETURN GP_ARG_FIRST
 #define FP_RETURN (UNITS_PER_FP_ARG == 0 ? GP_RETURN : FP_ARG_FIRST)
 
-#define MAX_ARGS_IN_REGISTERS 8
+#define MAX_ARGS_IN_REGISTERS (TARGET_RVE ? 6 : 8)
```
palmer-dabbelt: Do you mind submitting a patch to https://github.com/riscv/riscv-elf-psabi-doc/blob/master/riscv-elf.md that describes the E calling convention?

libgcc/config/riscv/multi3.S
```diff
@@ -79,3 +80,4 @@ __multi3:
   mv  a0, t2
   mv  a1, t4
   jr  t0
+#endif
```
palmer-dabbelt: Is there a generic version of this, or is it a FIXME?

gcc/common/config/riscv/riscv-common.c
```diff
@@ -77,6 +77,26 @@ riscv_parse_arch_string (const char *isa, int *flags, location_t loc)
...
     }
+  else if (*p == 'e')
+    {
+      p++;
+
+      *flags |= MASK_RVE;
+
+      if (*flags & MASK_64BIT)
+	{
+	  error ("E extension only support for RV32");
```
aswaterman: "E" is a base ISA, not an extension; maybe write "RV64E is not a valid base ISA"?

gcc/config.gcc
```diff
@@ -3991,7 +3991,7 @@ case "${target}" in
 			esac
 			;;
 		*)
-			echo "--with-arch=${with_arch} is not supported.  The argument must begin with rv32i, rv32g, rv64i, or rv64g." 1>&2
+			echo "--with-arch=${with_arch} is not supported.  The argument must begin with rv32e*, rv32i, rv32g, rv64i, or rv64g." 1>&2
```
aswaterman: remove the *

gcc/config/riscv/riscv.c
```diff
@@ -3085,7 +3085,16 @@ riscv_compute_frame_info (void)
...
-	frame->save_libcall_adjustment = x_save_size;
+	{
+	  frame->save_libcall_adjustment = x_save_size;
+	  if (TARGET_RVE && riscv_use_save_libcall (frame))
+	    {
+	      /* Libcall save/restore 3 registers at once, so we need
+		 allocate 12 bytes for callee-saved register.  */
+	      x_save_size = 12;
+	      frame->save_libcall_adjustment = x_save_size;
+	    }
```
aswaterman: I think this is clearer:

 	  if (TARGET_RVE && riscv_use_save_libcall (frame))
 	    {
 	      /* Libcall saves/restores 3 registers at once, so we need
 		 allocate 12 bytes for callee-saved register.  */
 	      x_save_size = 3 * UNITS_PER_WORD;
 	    }
	  frame->save_libcall_adjustment = x_save_size;

gcc/config/riscv/riscv.h
aswaterman: For RVE, the stack alignment requirement is just 4 bytes.  So, this is just a performance tradeoff (should I align the stack to 16 bytes so that I can compress ADDI into C.ADDI16SP).

Fortunately, RVE stack frames should be small (there are only 3 callee-saved registers: s0, s1, ra), so usually you can use C.ADDI to adjust the stack pointer.

aswaterman: I think this is OK

gcc/config/riscv/riscv.h
```diff
@@ -569,8 +569,11 @@ typedef struct {
 
 #define EPILOGUE_USES(REGNO)	((REGNO) == RETURN_ADDR_REGNUM)
 
-/* ABI requires 16-byte alignment, even on RV32. */
-#define RISCV_STACK_ALIGN(LOC) (((LOC) + 15) & -16)
+/* ABI requires 16-byte alignment, even on RV32,
```
aswaterman: May I suggest:

    #define RISCV_STACK_ALIGN(LOC) \ 
      (((LOC) + (BIGGEST_ALIGNMENT / 8)-1) & -(BIGGEST_ALIGNMENT / 8))

libgcc/config/riscv/multi3.S
aswaterman: I suggest that for RV32E, we rewrite this in C code, and inline the call to __mulsi3, so that RA doesn't need to be spilled.

libgcc/config/riscv/save-restore.S
```diff
@@ -294,6 +294,40 @@ __riscv_restore_0:
...
+  .cfi_restore 9
+  lw s0, 8(sp)
+  .cfi_restore 8
+  lw ra, 12(sp)
+  .cfi_restore 1
+  addi sp, sp, 16
+  .cfi_def_cfa_offset 0
+  ret
+  .cfi_endproc
+#else
```
aswaterman: Since RVE only has a 4-byte stack alignment, and we are only saving 3 registers, I think we should only adjust SP by 12 bytes, not 16.

palmer-dabbelt:
I forgot we hadn't merged this yet.  Is there anything still left to do aside from clean up the commits?

kito-cheng:
muldi for RV32E[1] under test, I guess I can finish it in next few day.

[1] https://github.com/riscv/riscv-gcc/pull/77/commits/040fef6f227f1fb6810dfe9bf1b175bd2c2697b0

palmer-dabbelt:
OK, no rush -- I just wanted to make sure you weren't waiting on us to do something.

libgcc/config/riscv/save-restore.S
kito-cheng: Done.

gcc/config/riscv/riscv.h
kito-cheng: Done.

gcc/config/riscv/riscv.c
kito-cheng: Done.

gcc/config.gcc
kito-cheng: Done.

gcc/common/config/riscv/riscv-common.c
kito-cheng: Done.

libgcc/config/riscv/multi3.S
kito-cheng: I've rewritten a multi3.S to a C version from libgcc2.c, it's also improve muldi3 for RV32 and multi3 for RV64.

kito-cheng:
This PR was split into 3 patch:
- **Implement __umulsidi3, umul_ppmm and __muluw3 for RISC-V** 
First patch implement __umulsidi3 and umul_ppmm, it leverage the fact of mulsi3/muldi3 only clobber`a0`, `a1`, `a2` and `ra`, and also improved performance of soft float.
- **Replace multi3.S with C version implementation**
Second patch is rewritten multi3.S, it's most grab from libgcc2.c but.
- Support RV32E
The main part of RV32E.

kito-cheng:
@palmer-dabbelt @aswaterman ping~

palmer-dabbelt:
Sorry, I'll try to look at this either tonight or tomorrow.

kito-cheng:
Update for fix some polluted to ChangeLog by accident :P

palmer-dabbelt:
Thanks, Kito!  I need to clean some stuff up in GCC land, but then I'll go submit these patches (and all the other cruft we have floating around) upstream.

asb:
If it's not already too late, _please_ don't merge upstream until the details of the RV32E ABI are properly documented in the psabi docs https://github.com/riscv/riscv-elf-psabi-doc/blob/master/riscv-elf.md

aswaterman:
Indeed it's not and let's keep it that way. @palmer-dabbelt 

The RV32EC ISA isn't finalized yet, either.

PR #88, opened 2017-08-16
Set SLOW_BYTE_ACCESS=1

aswaterman:
https://gcc.gnu.org/ml/gcc/2017-08/msg00202.html

We should make sure this doesn't regress on the test suite before sending upstream.

FYI @michaeljclark

### DIFF
--- a/gcc/config/riscv/riscv.h
+++ b/gcc/config/riscv/riscv.h
@@ -656,7 +656,7 @@ typedef struct {
 #define MOVE_MAX UNITS_PER_WORD
 #define MAX_MOVE_MAX 8
 
-#define SLOW_BYTE_ACCESS 0
+#define SLOW_BYTE_ACCESS 1
 
 #define SHIFT_COUNT_TRUNCATED 1
 


aswaterman:
No regressions.

aswaterman:
On rv64gc linux, that is.  I think that is sufficient for this sort of change.

michaeljclark:
That's good!

michaeljclark:
Yeah, it seems relatively safe.

palmer-dabbelt:
Can you make sure the strict align stuff still does the right thing?  IIRC this got flipped when we were making that change, righT?

aswaterman:
@palmer-dabbelt that was SLOW_UNALIGNED_ACCESS.  So I think this is OK.

michaeljclark:
I think it should be orthogonal and should honor alignment, whereas relaxing the alignment restriction was a less safe change. e.g. required adding -mstrict-align to linux-kernel if running under a monitor that isn't compliant. Given RISC-V allows unaligned accesses, it should be fine. Making it stricter is always going to be safer, as we know...

BTW I noticed that arm does their alignment fixups in linux-kernel, not in a monitor. I guess it just has to be done somewhere to be spec compliant.

- https://www.kernel.org/doc/Documentation/arm/mem_alignment

But this whole issue with unaligned accesses, just because they should be fixed up i.e. allowed, doesn't mean we should not take measures to avoid them until we are at a point where there is any system that handles them well. Even on x86 they are avoided. macOS compiler-rt requires 16 byte alignment for memory allocations due to the use of SSE.

I would actually like another change as I believe the choice to relax alignment for -Os is not a good idea. It should not be code size at all costs, and in this case, I don't think unaligned accesses actually will help a lot with code size. Unaligned accesses are rare (read the comment in the linux docs above) and -Os shouldn't be punished.

I think i'm going to patch my musl gcc distro like this:

```
--- gcc/config/riscv/riscv.c.orig	2017-08-17 12:44:49.000000000 +1200
+++ gcc/config/riscv/riscv.c	2017-08-17 12:43:24.000000000 +1200
@@ -312,21 +312,21 @@
 /* Costs to use when optimizing for size.  */
 static const struct riscv_tune_info optimize_size_tune_info = {
   {COSTS_N_INSNS (1), COSTS_N_INSNS (1)},	/* fp_add */
   {COSTS_N_INSNS (1), COSTS_N_INSNS (1)},	/* fp_mul */
   {COSTS_N_INSNS (1), COSTS_N_INSNS (1)},	/* fp_div */
   {COSTS_N_INSNS (1), COSTS_N_INSNS (1)},	/* int_mul */
   {COSTS_N_INSNS (1), COSTS_N_INSNS (1)},	/* int_div */
   1,						/* issue_rate */
   1,						/* branch_cost */
   2,						/* memory_cost */
-  false,					/* slow_unaligned_access */
+  true,					/* slow_unaligned_access */
 };
 
 /* A table describing all the processors GCC knows about.  */
 static const struct riscv_cpu_info riscv_cpu_info_table[] = {
   { "rocket", &rocket_tune_info },
   { "size", &optimize_size_tune_info },
 }; 
```

michaeljclark:
just wanted to rant. I'd like -Os to be relatively fast. it's all good 👍

michaeljclark:
Also trying to nut out how to get gcc to prefer RVC registers with -Os.

I haven't compile tested this yet... but I saw something similar in other arches. Just making a note of my findings here.

```
--- gcc/config/riscv/riscv.h.orig	2017-08-17 12:43:51.000000000 +1200
+++ gcc/config/riscv/riscv.h	2017-08-17 13:07:38.000000000 +1200
@@ -376,6 +376,7 @@
   SIBCALL_REGS,			/* registers used by indirect sibcalls */
   JALR_REGS,			/* registers used by indirect calls */
   GR_REGS,			/* integer registers */
+  RVC_REGS,			/* compressed extension accessible registers */
   FP_REGS,			/* floating-point registers */
   FRAME_REGS,			/* arg pointer and frame pointer */
   ALL_REGS,			/* all registers */
@@ -396,6 +397,7 @@
   "SIBCALL_REGS",							\
   "JALR_REGS",								\
   "GR_REGS",								\
+  "RVC_REGS",								\
   "FP_REGS",								\
   "FRAME_REGS",								\
   "ALL_REGS"								\
@@ -418,6 +420,7 @@
   { 0xf00000c0, 0x00000000, 0x00000000 },	/* SIBCALL_REGS */	\
   { 0xffffffc0, 0x00000000, 0x00000000 },	/* JALR_REGS */		\
   { 0xffffffff, 0x00000000, 0x00000000 },	/* GR_REGS */		\
+  { 0x0000ff00, 0x00000000, 0x00000000 },	/* RVC_REGS */		\
   { 0x00000000, 0xffffffff, 0x00000000 },	/* FP_REGS */		\
   { 0x00000000, 0x00000000, 0x00000003 },	/* FRAME_REGS */	\
   { 0xffffffff, 0xffffffff, 0x00000003 }	/* ALL_REGS */		\
--- gcc/config/riscv/riscv.c.orig	2017-08-17 12:44:49.000000000 +1200
+++ gcc/config/riscv/riscv.c	2017-08-17 13:10:15.000000000 +1200
@@ -612,6 +612,17 @@
   return riscv_const_insns (x) > 0;
 }
 
+/* Implement TARGET_PREFERRED_RELOAD_CLASS. */
+
+static reg_class_t
+riscv_preferred_reload_class (rtx x ATTRIBUTE_UNUSED, reg_class_t rclass)
+{
+  if (rclass == GENERAL_REGS && TARGET_RVC && optimize_size)
+    return RVC_REGS;
+  else
+    return rclass;
+}
+
 /* Implement TARGET_CANNOT_FORCE_CONST_MEM.  */
 
 static bool
@@ -4098,6 +4109,9 @@
 #define TARGET_HAVE_TLS true
 #endif
 
+#undef  TARGET_PREFERRED_RELOAD_CLASS
+#define TARGET_PREFERRED_RELOAD_CLASS riscv_preferred_reload_class
+
 #undef TARGET_CANNOT_FORCE_CONST_MEM
 #define TARGET_CANNOT_FORCE_CONST_MEM riscv_cannot_force_const_mem
 ```

aswaterman:
That only happens if you do -Os -mtune=size.  If you just do -Os it shouldn't do misaligned accesses.

michaeljclark:
I'm compile testing the RVC register patch (against gcc-7.2.0) and I just noticed that riscv_tune_info doesn't have the slow_unaligned_access field in gcc-7.2.0 (as I got a reject when backporting my patch) so this is a new feature?

Is there a good improvement in size? Is the size improvement worth it for the expense in terms of performance? How do we strike a balance between size and speed? Apple's approach to -Os is size  but not at the expense of speed (which now manifests in Clang):

- https://gcc.gnu.org/ml/gcc/2017-08/msg00006.html

I actually dug out apple-gcc last night so I could diff it against FSF gcc. It's gcc-4.2.1. I was curious about an -Os that is good for size and speed, not size at all costs.

SLOW_BYTE_ACCESS=1 will improve size and performance, as size and speed can often be correlated.

michaeljclark:
OK. I misunderstood the patch.

michaeljclark:
Is optimize_size set for -Os? i.e. wondering if the RVC patch above will work.... compiling now...

aswaterman:
It already favors RVC regs using ```riscv_register_priority``` -- are you sure the change you've proposed will help additionally (and not hurt for other reasons, e.g., increased spills)?

michaeljclark:
The additional patch is just an experiment at this point. I saw CHEAP_REGISTERS in arc, and something similar in SPARC. It's not an obvious change like this one, but worth trying.

michaeljclark:
OK. I see riscv_register_priority now. I missed that.

I wonder how the priority value is factored into the cost. I see i386 uses 2, 3 and 4. Not just 1.

michaeljclark:
The hope is this doesn't create more spills, just prefers RVC registers for reloads.

Perhaps riscv_register_priority already does this. sN registers could potentially be less favorable in lead functions than aN and tN because they require saving and restoring. e.g. 0, 1 and 2 rather than just 0 and 1. We have to do some analysis on how well the register allocator works, and how it responds to hints...

However when I did histograms on dynamic register usage with gcc-7.1 and -Os I saw non-RVC registers with higher dynamic frequencies than RVC registers. It didn't appear to strongly favour RVC registers. It seemed to favor RVC registers less than gcc-6.3, which was my previous baseline, however, this is anecdotal until we spend more time looking at results...

michaeljclark:
The `riscv_preferred_reload_class` patch does not help, although I would like to study what effect it had on the register allocation, to understand how the register allocator uses the preferred reload class.

I might test returning different values from `riscv_register_priority`.

kito-cheng:
Just provide another directions for optimize code size: `TARGET_PREFERRED_RENAME_CLASS`[1] or `preferred_for_size`[2], but I think it require lots work in `riscv.md` and `riscv.h` if you want to try `preferred_for_size` :P

[1] https://gcc.gnu.org/onlinedocs/gccint/Register-Classes.html
[2] https://gcc.gnu.org/onlinedocs/gccint/Disable-Insn-Alternatives.html

michaeljclark:
@kito-cheng thanks!

It seems that TARGET_PREFERRED_RENAME_CLASS is more appropriate than TARGET_PREFERRED_RELOAD_CLASS as it mentions ARM, thumb-2 instructions using LO_REGS may be smaller than instructions using GENERIC_REGS.

I'm going to run some more experiments locally...

PR #91, opened 2017-09-01
RISC-V: Ignore Custom Extensions

palmer-dabbelt:
The RISC-V ISA allows custom ISA extensions to be defined by users.
These extensions must come after all the standard extensions, and must
start with 'x'.  This patch allows these custom extensions to be passed
via the '-march' flag and ignores them -- we don't plan on supporting
any custom extensions in GCC, so I think that's always the right thing
to do.  These extensions will be passed to the assembler, which is
expected to have either been modified to support the extension or
produce an error.

gcc/ChangeLog

2017-09-01  Palmer Dabbelt  <palmer@dabbelt.com>
            Joel Vandergriendt <vandergriendtjoel@gmail.com>

        * common/config/riscv/riscv-common.c (riscv_parse_arch_string):
        Allow custom extensions via -march.

### DIFF
--- a/gcc/common/config/riscv/riscv-common.c
+++ b/gcc/common/config/riscv/riscv-common.c
@@ -87,7 +87,7 @@ riscv_parse_arch_string (const char *isa, int *flags, location_t loc)
   if (*p == 'c')
     *flags |= MASK_RVC, p++;
 
-  if (*p)
+  if (*p && *p != 'x')
     {
       error_at (loc, "-march=%s: unsupported ISA substring %qs", isa, p);
       return;


palmer-dabbelt:
@kito-cheng @aswaterman I think it's sane to just ignore custom extensions.  Unless anyone says anything I'll submit upstream.

aswaterman:
Seems sensible.

kito-cheng:
It was ignore `Xcustom` I remember, and now just ignore `x`, btw @aswaterman do you remember the full story why `Xcustom` is not supported?

aswaterman:
We didn't want to give special permission to one non-standard extension, because then we'd need a process for accepting or rejecting other non-standard extensions.  It was simpler to just reject them all.

@sorear had a proposal for how to deal with this in GAS, so at least we can support inline asm for things like Xcustom: https://github.com/riscv/riscv-binutils-gdb/issues/38

vanjoe:
Will this be merged?

palmer-dabbelt:
Yes, but I need to do a bit of GCC maintenance before I submit any patches upstream as it's been a bit since we've re-tested upstream's trunk.  Is this blocking you?  It's a pretty safe patch, so I don't mind accelerating things for it.

palmer-dabbelt:
Sorry, I thought you meant upstream -- I didn't notice we hadn't merged the PR.  Thanks for pinging :)

vanjoe:
Yeah, that's all I was curious about.

PR #93, opened 2017-09-07
For RV32I ABIs, stack alignment should only be 64 bits

aswaterman:
See https://github.com/riscv/riscv-elf-psabi-doc/blob/master/riscv-elf.md

### DIFF
--- a/gcc/config/riscv/riscv.h
+++ b/gcc/config/riscv/riscv.h
@@ -124,7 +124,7 @@ along with GCC; see the file COPYING3.  If not see
 #define FUNCTION_BOUNDARY (TARGET_RVC ? 16 : 32)
 
 /* There is no point aligning anything to a rounder boundary than this.  */
-#define BIGGEST_ALIGNMENT (TARGET_RVE ? 32 : 128)
+#define BIGGEST_ALIGNMENT (TARGET_RVE ? BITS_PER_WORD : (2 * BITS_PER_WORD))
 
 /* The user-level ISA permits unaligned accesses, but they are not required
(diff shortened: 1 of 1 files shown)


kito-cheng:
Does this change effect to code size on RV32I since c.addi16sp may less generated?

palmer-dabbelt:
We only require 4 byte alignment on RV32E already, so it's no change.

On Sep 7, 2017 8:17 PM, "Kito Cheng" <notifications@github.com> wrote:

Does this change effect to code size on RV32I since c.addi16sp may less
generated?

—
You are receiving this because you modified the open/close state.

Reply to this email directly, view it on GitHub
<https://github.com/riscv/riscv-gcc/pull/93#issuecomment-327988264>, or mute
the thread

kito-cheng:
Yeah, but RV32I change to 8 byte alignment, not multiple of 16?

palmer-dabbelt:
Sorry, that didn't make any sense.  I guess the point was: we're making the
same trade-off here as we do in RV32E, where we compress when possible, and
since it's a good idea there it should be here as well.

Relaxing the ABI just let's the compiler make more decisions, though you're
right that with how we do it now it'll compress less often.

aswaterman:
However, all stack frames <= 32 bytes will still compress. That handles
functions that save up to 8 registers or have up to 8 words of stack
storage. That should be the majority.

And of course 50% of the remaining cases still compress.

aswaterman:
BTW the compiler can still choose to over-align the stack to 16 bytes if it
wants :)

PR #95, opened 2017-09-08
Support -mpreferred-stack-boundary flag

aswaterman:
This supports reducing the stack alignment to 8 bytes for RV32I.

### DIFF
--- a/gcc/config/riscv/riscv.c
+++ b/gcc/config/riscv/riscv.c
@@ -217,6 +217,9 @@ struct riscv_cpu_info {
 /* Whether unaligned accesses execute very slowly.  */
 bool riscv_slow_unaligned_access;
 
+/* Stack alignment to assume/maintain.  */
+unsigned riscv_stack_boundary;
+
 /* Which tuning parameters to use.  */
 static const struct riscv_tune_info *tune_info;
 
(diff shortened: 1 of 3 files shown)


kito-cheng:
I think it's a disaster if mixing objects with different stack alignment, so it would be better if record stack alignment in object and let linker check the consistency between objects?

aswaterman:
Yeah, sticking this in the elf flags is probably a good idea.

palmer-dabbelt:
I'm going to merge this for now, we'll figure out the ELF flags stuff before submitting upstream.

kito-cheng:
Maybe we could put such info in  special sections in ELF instead of ELF flags since here is only 32 bits, and stack alignment can be various.

Furthermore thought is about the info for non-standard extension also can store by similar way too.

palmer-dabbelt:
At some point we were kicking around the idea of having a RISC-V section (IIRC something like ".ARCH.riscv" is reserved for us) that contains some extra info like the ISA used to compile and such, there's just been no time to do it.

kito-cheng:
OK, I'll try to propose something about risc-v special sections next month on `riscv/riscv-elf-psabi-doc` and implement an initial implementation for binutils.

PR #98, opened 2017-09-24
Define MUSL_DYNAMIC_LINKER for riscv

michaeljclark:
This patch teaches gcc configured with a riscv*linux-musl tuple to correctly set the program interpreter for dynamically linked binaries. gcc already has support for musl-libc and just requires the port to define MUSL_DYNAMIC_LINKER.

This patch uses the same scheme as other ports such that the musl libc program interpreter can co-exist with glibc. i.e. `ld-musl-<arch>-<abi>.so.1` instead of `ld-linux-<arch>-<abi>.so.1`.

Tested with gcc 7.2 using the following bootstrap script and musl-riscv port:

- https://github.com/rv8-io/musl-riscv-toolchain
- https://github.com/lluixhi/musl-riscv

Verify program interpreter using readelf

```
$ readelf -l libc-test-bin/argv.exe 

Elf file type is EXEC (Executable file)
Entry point 0x10440
There are 6 program headers, starting at offset 64

Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  PHDR           0x0000000000000040 0x0000000000010040 0x0000000000010040
                 0x0000000000000150 0x0000000000000150  R E    0x8
  INTERP         0x0000000000000190 0x0000000000010190 0x0000000000010190
                 0x0000000000000020 0x0000000000000020  R      0x1
      [Requesting program interpreter: /lib/ld-musl-riscv64-lp64d.so.1]
  LOAD           0x0000000000000000 0x0000000000010000 0x0000000000010000
                 0x00000000000008cc 0x00000000000008cc  R E    0x1000
  LOAD           0x0000000000000e50 0x0000000000011e50 0x0000000000011e50
                 0x0000000000000200 0x0000000000000240  RW     0x1000
  DYNAMIC        0x0000000000000e60 0x0000000000011e60 0x0000000000011e60
                 0x00000000000001a0 0x00000000000001a0  RW     0x8
  GNU_RELRO      0x0000000000000e50 0x0000000000011e50 0x0000000000011e50
                 0x00000000000001b0 0x00000000000001b0  R      0x1

 Section to Segment mapping:
  Segment Sections...
   00     
   01     .interp 
   02     .interp .hash .dynsym .dynstr .rela.plt .plt .text .rodata .eh_frame 
   03     .init_array .fini_array .dynamic .got .sdata .sbss .bss 
   04     .dynamic 
   05     .init_array .fini_array .dynamic 
```

Testing a dynamic binary compiled with musl-riscv

```
/ # cd /lib
/lib # ls -l
lrwxrwxrwx    1 0        0                7 Sep 23  2017 ld-musl-riscv64-lp64d.so.1 -> libc.so
-rwxr-xr-x    1 0        0           796648 Sep 23  2017 libc.so
/lib # cd /tmp
/tmp # tar xf libc-test-bin.tar 
/tmp # cd libc-test-bin/
/tmp/libc-test-bin # ./argv.exe
/tmp/libc-test-bin # uname -a
Linux ucbvax 4.6.2-g250754b #4 Sun Sep 24 12:05:47 NZDT 2017 riscv64 GNU/Linux
```

### DIFF
--- a/gcc/config/riscv/linux.h
+++ b/gcc/config/riscv/linux.h
@@ -24,6 +24,17 @@ along with GCC; see the file COPYING3.  If not see
 
 #define GLIBC_DYNAMIC_LINKER "/lib/ld-linux-riscv" XLEN_SPEC "-" ABI_SPEC ".so.1"
 
+#define MUSL_ABI_SUFFIX \
+  "%{mabi=ilp32:-sf}" \
+  "%{mabi=ilp32f:-sp}" \
+  "%{mabi=ilp32d:}" \
+  "%{mabi=lp64:-sf}" \
+  "%{mabi=lp64f:-sp}" \
+  "%{mabi=lp64d:}" \
+
+#undef MUSL_DYNAMIC_LINKER
+#define MUSL_DYNAMIC_LINKER "/lib/ld-musl-riscv" XLEN_SPEC MUSL_ABI_SUFFIX ".so.1"
+
 /* Because RISC-V only has word-sized atomics, it requries libatomic where
    others do not.  So link libatomic by default, as needed.  */
 #undef LIB_SPEC


aswaterman:
lgtm, @palmer-dabbelt ?

michaeljclark:
We are now discussing the musl linker scheme on #musl at freenode...

Path of least resistance is to use the same scheme as glibc. Note the musl port has not gone upstream so perhaps we can wait on merging this... It's prompted some discussion...

aswaterman:
We discussed this to death for glibc and are pretty confident in our conclusion. It would be unfortunate for musl to do something different without a very compelling reason.

michaeljclark:
Summarising the discussion with the musl maintainers on IRC:

- riscv32-ilp32d has a redundant 32
- riscv64-lp64d has a redundant 64
- lp64d and ilp32d are the main "platform" variants and the lp64d and ilp32d can be omitted e.g.
  - ld-musl-riscv32.so.1
  - ld-musl-riscv64.so.1
- instead, distinguish the float double reg, float reg, double mem variants.
  - matches powerpc which has powerpc and powerpc-sf
- understand the current scheme allows an x32 variant e.g. riscv64-ilp32d
  - but this is not a default for the "platform" and so could also have a suffix

The comments:

- dalias: the viability of an arch as a platform is inversely proportional to number of abi variants
- dalias: 1. that it doesn't match how musl would treat rv64-ilp32 if we decide to support it
- dalias: 2. that it's full of gratuitous verbosity representing things that aren't variable
- dalias: re: glibc, i don't have a strong opinion, but i think changing it like what i'm proposing for musl makes for a more coherent, serious-looking "platform"

We would have to define the suffix for the non default ABI variants as there are the f,d variants. i.e. we can't just use -sf

interp | abi
:-- | :--
ld-musl-riscv32.so.1 | riscv32-ilp32d
ld-musl-riscv64.so.1 | riscv64-lp64d
ld-musl-riscv32-sf.so.1 | riscv32-ilp32
ld-musl-riscv64-sf.so.1 | riscv64-lp64
ld-musl-riscv32-hfsd.so.1 ? | riscv32-ilp32
ld-musl-riscv64-hfsd.so.1 ? | riscv64-lp64

The rationale is that these will be the most common and define the "riscv platform" for linux

interp | abi
:-- | :--
ld-musl-riscv32.so.1 | riscv32-ilp32d
ld-musl-riscv64.so.1 | riscv64-lp64d

and more unusual variants should have suffixes (hfsd is hard float soft double for want of a better suffix for this unusual variant); given we are talking about unix platforms that will likely have float emulation.

michaeljclark:
I edited that last comment a couple of times, so if you got email notification, see the pull request

michaeljclark:
i noted there would be resistance to changing the scheme for musl out of the argument for consistency, and in fact I'm fine with the current scheme, out of consistency; but also understand the rationale for simplification of the main platform variant.

michaeljclark:
I've edited the comment with feedback from sorear. The only different is in the main platform variant, that we omit the suffix. It is indeed consistent with the current scheme but omits the additional tuple value for the main unix platform variant.

Hope you are okay with that and I can spin another patch.

michaeljclark:
@sorear

michaeljclark:
If you decided to use this on glibc it could be trivially supported and backward compatible by adding symlinks. fedora for example would only be using the d variants and symlinks in the glibc install/rpm would prevent them from spitting the dummy if this was changed. I'm only suggesting that.

however for musl the scheme is as yet undefined.

have a look at other platforms that define MUSL_DYNAMIC_LINKER. musl has its own scheme in some ports and doesn't necessarily always follow glibc for various reasons. it is kind of something that is musl defined.

sorear:
The other thing @richfelker brought up is that requiring float registers + FMV.[SD] + F[LS][WD] in the POSIX platform would improve the ABI situation.

aswaterman:
We are not changing glibc.

We strongly feel the different C libraries should adhere to the same naming conventions.  It sounds like your concerns are merely aesthetic.  Why differ on that basis alone?  And what are you going to do about riscv64-ilp32?

aswaterman:
Mandating floating-point isn't realistic.  Vendors are going to create hardware that doesn't have it, and if the various unix ports don't support it, they will create their own.  This will cause more fragmentation, not less.

michaeljclark:
I'm agnostic to the suffix, however, the full linux systems are likely to have float and double. I understand Fedora is only going to do RV64GC iirc. Certainly embedded systems with shared libraries may have soft float as did earlier arm systems, although hard float is standard on Raspberry Pi class boards.

I've prepared an alternate patch to show the musl maintainers but I haven't pushed it. https://hastebin.com/jesixabaca.cpp

michaeljclark:
I think the point from the musl maintainers if that we need to distinguish embedded vs the linux platform here. The distros are not going to support that many abi variants as if effectively means a separate port for each abi. arm was a real mess. Debian has arm and armhf. I don't think folk like the suffix given armhf is what people typically use. RISC-V could set the bar at IMAFDC for the "unix platform".

https://wiki.debian.org/ArmHardFloatPort

richfelker:
I think I (musl maintainer) should have just replied here right away rather than having results of our discussion yesterday go through others. That's because the discussion was complex and a mix of ideals ("wouldn't it be better if things were this way: ...?") and real concerns within the scope of what can and can't be changed. For example I would love it if riscv always had fp registers even when it doesn't have insns to operate on them, so that there's no need for different ABIs, but that's not something I can change/dictate.

Anyway, back to the points of my objection to the proposed ldso naming scheme:

1. Impedance mismatch with musl's arch system, which is what the musl ldso-name is for.

musl's convention across all archs is for the ldso name to be ld-musl-$(ARCH).so.1 for the base arch (default ABI that's intended as what users of "normal" Linux systems will be using) and ld-musl-$(ARCH)$(SUBARCH).so.1 for alternate ABI variants. In our scheme of archs, minor variants like LE/BE, soft/hard float, etc. are subarchs while differences significant enough to affect basic type definitions/sizes (word/pointer size) or different sets of syscalls are treated as fully separate archs. For example, i386, x86_64, and x32 are all separate archs and mips, mips64, and mipsn32 are all separate archs. If we at some point want/need to support riscv64-ilp32, it would be a separate top-level arch, not subarch. The proposed naming on the other hand suggests it's a subarch of riscv64. (I say suggests because the full "riscv64-lp64" and "riscv64-ilp32" names *could* be treated as top-level archs, but that requires renaming everything in the existing proposed port and I think it's ugly and inappropriate.)

2. Gratuitous naming clutter.

Especially for riscv32, having "-ilp32" on the end adds no meaning at all. There is nothing it could possibly be but ilp32. For riscv64, it's more a matter of the above principles we use for naming archs; the "normal" ABI should just be an unadorned base arch name. This is not only "prettier" but it alleviates an end-user problem where they run readelf/file(1) on a binary to check the PT_INTERP (or ls /lib), see a big string of ABI combinatorics, and have to do research to determine if the binary is a "normal $ARCH binary" or some specialized ABI variant. If the default/normal ABI is named in a way that it's clearly the default, then most of the time the answer is immediately clear.

Now, where there still is an open question is what do to about float ABIs. My understanding is that the long-term intent for serious Linux systems is to be running with full hard-float, and if so my leaning is to have the ABI for the unadorned "riscv32" and "riscv64" archs be full hard-float (single and double in registers) with softfloat and perhaps (if there's demand?) hard-single-only/soft-double as ABI variants. For example on mips and powerpc where hard-float is default, softfloat ABI variants are named with "-sf" in the subarch. SH calls this "-nofpu" for historical reasons. Both SH and powerpc are actually cases where there's some additional complexity in that the tools (GCC) support a few weird models with single-precision only or other abnormal FPUs, but at this time we just don't support those ABI variants at all. Whether to have musl support single-only on riscv probably depends on whether there will be interesting hard cpus with this limitation that people want to use, but it doesn't hurt to reserve ABI naming at this time anyway. I don't have a strong preference for what the naming should be, although "-sf" suffix would at least be consistent with how we do most other archs (so "riscv32-sf" and "riscv64-sf") and maybe "-sd" for "soft double". Other suggestions are welcome.

OTOH if it really is the intent from the riscv side that soft float be the default (this seems doubtful to me from the fact that G profile includes F and D), it probably makes sense to do the same on the musl naming side.

aswaterman:
Thanks @richfelker for the explanation of the musl conventions.  Treating rv64+ilp32 as a different top-level arch makes sense and addresses my principal objection: in this scheme, the extra set of 32's or 64's is truly redundant, so dropping them makes sense.

It is indeed my expectation that rv32+ilp32d and rv64+lp64d will be the standard choice for general-purpose Linux systems.

@michaeljclark can you update this PR to reflect the desired naming convention?

michaeljclark:
I'm compiling and testing a new patch...

richfelker:
I still think using the suffix "-lp64" to mean "softfloat ABI" is really confusing/misleading.

michaeljclark:
I've defined MUSL_ABI_SUFFIX in gcc/config/riscv/linux.h as I did not change the start file lookup directory which still uses the current ABI_SPEC definition. I used SUFFIX instead of SPEC because it is sometimes empty.

```
$ readelf -l hello

Elf file type is EXEC (Executable file)
Entry point 0x103c8
There are 6 program headers, starting at offset 64

Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  PHDR           0x0000000000000040 0x0000000000010040 0x0000000000010040
                 0x0000000000000150 0x0000000000000150  R E    0x8
  INTERP         0x0000000000000190 0x0000000000010190 0x0000000000010190
                 0x000000000000001a 0x000000000000001a  R      0x1
      [Requesting program interpreter: /lib/ld-musl-riscv64.so.1]
  LOAD           0x0000000000000000 0x0000000000010000 0x0000000000010000
                 0x00000000000004d4 0x00000000000004d4  R E    0x1000
  LOAD           0x0000000000000e50 0x0000000000011e50 0x0000000000011e50
                 0x00000000000001f0 0x0000000000000228  RW     0x1000
  DYNAMIC        0x0000000000000e60 0x0000000000011e60 0x0000000000011e60
                 0x00000000000001a0 0x00000000000001a0  RW     0x8
  GNU_RELRO      0x0000000000000e50 0x0000000000011e50 0x0000000000011e50
                 0x00000000000001b0 0x00000000000001b0  R      0x1

 Section to Segment mapping:
  Segment Sections...
   00     
   01     .interp 
   02     .interp .hash .dynsym .dynstr .rela.plt .plt .text .rodata .eh_frame 
   03     .init_array .fini_array .dynamic .got .sdata .bss 
   04     .dynamic 
   05     .init_array .fini_array .dynamic 
```

aswaterman:
Using the suffix -lp64 to mean softfloat is confusing because of not using the suffix -lp64d to mean hardfloat.  If musl deviates by omitting the -lp64d suffix, it seems fine to deviate on the other suffixes, too...

michaeljclark:
@richfelker what do you propose? given we have two softfloat ABIs. well one semi-soft (hard SP, soft DP) ;-)

I followed this approach in response to @sorear mentioning to make the other suffixes consistent with the current ABI_SPEC that is used for the start files directory.

michaeljclark:
This actually raises the point mentioned on the mailing list about -mnewlib, -mmusl, -mglibc and whether the compiler can have separate start file directories depending on the std lib.

However I don't want to make this current pull request to complex. Currently we get /dev/null for the program interpreter when compiling dynamic libraries when riscv-gcc is configured for a riscv*linux-musl tuple

michaeljclark:
s/to complex/too complex/

michaeljclark:
how about this? (sf = soft float, sp = single precision)

interp | abi
:-- | :--
ld-musl-riscv32.so.1 | riscv32-ilp32d
ld-musl-riscv64.so.1 | riscv64-lp64d
ld-musl-riscv32-sf.so.1 | riscv32-ilp32
ld-musl-riscv64-sf.so.1 | riscv64-lp64
ld-musl-riscv32-sp.so.1 | riscv32-ilp32f
ld-musl-riscv64-sp.so.1 | riscv64-lp64f

michaeljclark:
riscv64-lp64f is likely to be rare. riscv32-ilp32f may exist with implementations that have a shared 32-bit physical register file. they need a name. -sp and -sf seem okay?

michaeljclark:
I can squash the commits to make a clean patch with documentation for the scheme in the commit. I'm not sure if I push a new ref with these three commits squashed or whether I need to open a new PR... @richfelker  lemme know if you are fine with -sp for the hard single, soft double ABI. It's likely not to be a common configuration.

richfelker:
Those all look fine to me. Thanks!

aswaterman:
OK, I'll leave it to @palmer-dabbelt from here

michaeljclark:
Cool. Here's a description if we decide to squash the commit. 

```
Define MUSL_DYNAMIC_LINKER for riscv                      

Use no suffix at all in the musl dynamic linker name for hard
float ABI. Use -sf and -sp suffixes in musl dynamic linker name
for soft float and single precision ABIs. The following table
outlines the musl interpreter names for the RISC-V ABI names.

musl interpreter        | RISC-V ABI
----------------------- | -------------
ld-musl-riscv32.so.1    | riscv32-ilp32d
ld-musl-riscv64.so.1    | riscv64-lp64d
ld-musl-riscv32-sf.so.1 | riscv32-ilp32
ld-musl-riscv64-sf.so.1 | riscv64-lp64
ld-musl-riscv32-sp.so.1 | riscv32-ilp32f
ld-musl-riscv64-sp.so.1 | riscv64-lp64f
```

@palmer-dabbelt lemme know if you want me to squash so we have a clean history. I actually don't know what happens to a GitHub PR if I squash and push -f to the branch on this PR.

michaeljclark:
Just noticed the ld.so name now matches the musl install scripts which defaults to the architecture name (which is good news):

```
./tools/install.sh -D -m 755 lib/libc.so /opt/riscv/musl-riscv-toolchain-7.2.0-4/riscv64-linux-musl/lib/libc.so
./tools/install.sh -D -l /lib/libc.so /opt/riscv/musl-riscv-toolchain-7.2.0-4/riscv64-linux-musl/lib/ld-musl-riscv64.so.1 || true
```

where DESTDIR, in this case, is set to the sysroot for the triple

`DESTDIR=/opt/riscv/musl-riscv-toolchain-7.2.0-4/riscv64-linux-musl/`

Note: when creating the sysroot, if you append the target triple to the install dir, you can nicely support multiple toolchains with one install prefix and the sysroot matches the internal gcc scheme. I noticed riscv-gnu-toolchain uses $(INSTALL_DIR)/sysroot which prevents multiple toolchains from being installed to the same prefix.

We've now got a musl-riscv toolchain with working pthreads/atomics (we still need to get the toolchain building on macos, as the mac toolchain `ar` won't create empty archives, so we'll need to patch musl-libc to add an empty object to the compatibility libs):

https://github.com/rv8-io/musl-riscv-toolchain/releases/tag/v7.2.0-4

palmer-dabbelt:
@michaeljclark did you guys decide what's going on here?  I'm OK taking the diff that's there currently.

michaeljclark:
Hi @palmer-dabbelt, I think it's fine to merge as is but if you want me to I can squash the 3 commits with one log message describing the scheme (as it evolved). I guess it depends on how you are going to send the patch(es) upstream. If I squash it and push -f; I suspect GitHub will figure out that the tip of the branch has changed, and you can then just merge one commit. Up to you...

palmer-dabbelt:
If you squash then github will figure it out.

On Sep 28, 2017 3:29 PM, "Michael Clark" <notifications@github.com> wrote:

Hi @palmer-dabbelt <https://github.com/palmer-dabbelt>, I think it's fine
to merge as is but if you want me to I can squash the 3 commits with one
log message describing the scheme (as it evolved). I guess it depends on
how you are going to send the patch(es) upstream. If I squash it and push
-f; I suspect GitHub will figure out that the tip of the branch has
changed, and you can then just merge one commit. Up to you...

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/riscv/riscv-gcc/pull/98#issuecomment-332981331>, or mute
the thread

michaeljclark:
ok...

michaeljclark:
OK. I've flattened it and given it a nice log message...

michaeljclark:
@richfelker was fine with it from the musl side of things... the default interpreter now matches where the musl Makefile installs ld.so to e.g. /lib/ld-musl-riscv64.so.1 for riscv64-lp64d.

michaeljclark:
@palmer-dabbelt  Just browsing the commit log. I noticed the strict-operands patch is on `riscv-next` but not `riscv-gcc-7`.

Do you need to cherry-pick the musl dynamic loader commit onto `riscv-next`?

I'm not sure which branch I should request merges to in future... I suppose `riscv-next` will eventually sync with `gcc-8` and `riscv-gcc-7` will rebase against gcc-7 for maintenance patches. e.g. changes that might go into gcc-7.3? In any case you might want to apply this patch to `riscv-next` so it's also on the `gcc-8` train.

palmer-dabbelt:
Sorry, it's a bit messy right now, partially because we've moving things around a bit.  The idea is that:

* There will be a "riscv-gcc-7.2.0" branch that has backported patches onto the 7.2.0 release.
* The "riscv-gcc-7" branch will be deprecated, as it is rebased onto the upstream release branch and that's a bit of a pain.
* "riscv-next" will remain the queue.

The flow will look like "riscv-next" -> "trunk" -> "riscv-gcc-7.2.0".  The reason this is taking a while is that I'm going through each patch (starting with binutils) and making sure it's good enough for upstream, committing it upstream, and then backporting it.  This is how we're supposed to do it, there was just a bit of a break down in the process due to all the craziness over the summer.

michaeljclark:
OK No worries. Next time I'll queue into `riscv-next`. This is a pretty simple and discrete patch so it shouldn't make things too hard. Cheers, Michael.

PR #103, opened 2017-10-30
Allow use of fs0-fs11 as call-clobbered registers in soft-float ABI

aswaterman:
Note this change does not cause ABI breakage, because previously we had disallowed use of fs0-fs11 altogether in the soft-float ABI.


aswaterman:
sgtm

PR #105, opened 2017-11-03
Mark integer library routines as functions

PkmX:
Currently you can't link those in `libgcc_s.so`:

```c
unsigned udivsi3(unsigned a, unsigned b) {
    return a / b;
}

int main(void) {
    return 0;
}
```

```
$ riscv32-unknown-linux-gnu-gcc -march=rv32i -shared-libgcc main.c
ld: warning: type and size of dynamic symbol `__udivsi3@@GCC_3.0' are not defined
```

This patch marks them as functions so the linker knows that it should create a PLT entry.


aswaterman:
thanks, @palmer-dabbelt can you send it upstream

PR #106, opened 2017-11-07
RISC-V: Define _REENTRANT if -pthread is given

kito-cheng:
- AX_PTHREAD (provide by ax_pthread.m4) will check this marco is
   define or not for check pthread is available or not.
   - https://savannah.gnu.org/patch/?8186
   - https://www.gnu.org/software/autoconf-archive/ax_pthread.html

 - Most of all other target will do the same thing in their linux
   port, e.g. ARM, AArch64, i386, mips...


palmer-dabbelt:
Thanks!

PR #110, opened 2017-11-20
Add -mabi=ilp32e

kito-cheng:
https://github.com/riscv/riscv-toolchain-conventions/issues/3

### DIFF
--- a/gcc/config.gcc
+++ b/gcc/config.gcc
@@ -4041,6 +4041,7 @@ case "${target}" in
 		"")
 			# Infer XLEN, but otherwise assume GC.
 			case "${with_abi}" in
+			ilp32e) with_arch="rv32e" ;;
 			ilp32 | ilp32f | ilp32d) with_arch="rv32gc" ;;
 			lp64 | lp64f | lp64d) with_arch="rv64gc" ;;
 			*) with_arch="rv${xlen}gc" ;;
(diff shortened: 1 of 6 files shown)


palmer-dabbelt:
Thanks, Kito.  I probably won't have time to look at this for a bit, but I appreciate the help :)

gcc/config.gcc
```diff
@@ -4056,11 +4057,12 @@ case "${target}" in
 		# pick a default based on the ISA, preferring soft-float
 		# unless the D extension is present.
 		case "${with_abi}" in
-		ilp32 | ilp32f | ilp32d | lp64 | lp64f | lp64d)
+		ilp32 | ilp32f | ilp32d | lp64 | lp64f | lp64d | lp64q)
```
jim-wilson: ilp32e needs to be in this list of valid abis, otherwise you are overriding the user's choice, and won't give an error if the user makes a mistake with non-matching arch/abi.

gcc/config/riscv/riscv.c
```diff
@@ -4025,12 +4028,14 @@ riscv_conditional_register_usage (void)
...
-	  call_used_regs[r] = call_really_used_regs[r] = 1;
-	}
+      for (int r = 16; r <= 31; r++)
+	fixed_regs[r] = 1;
+    }
+
+  if (riscv_abi == ABI_ILP32E)
+    {
+      for (int r = 16; r <= 31; r++)
+	call_used_regs[r]  = 1;
```
jim-wilson: An extra space before the =, minor nit.

gcc/config.gcc
kito-cheng: Fixed, that's my mistake when rebase with Q-extension...

gcc/config/riscv/riscv.c
kito-cheng: Fixe :)

kito-cheng:
@palmer-dabbelt 
We have Jim now!

@jim-wilson 
ABI_SPEC also fixed. And linux kernel doesn't support rv32e so just keep MUSL_ABI_SUFFIX unchanged.

thanks for your review :)

jim-wilson:
I am unfamiliar with github workflow in general, and riscv github workflow in particular.  I was assuming that the person who submitted the patch would merge them after approval.  This is how FSF binutils/gcc work.  If you are waiting for me to merge them, let me know.

PR #113, opened 2017-12-14
Use __riscv_flush_icache to flush the I$ on Linux

palmer-dabbelt:
The fence.i instruction is not appropriate, because it only flushes
the current core's I$.  If the thread is migrated to a different core,
the instructions would be incoherent.

This depends on https://github.com/riscv/riscv-glibc/pull/22

### DIFF
--- a/gcc/config/riscv/linux.h
+++ b/gcc/config/riscv/linux.h
@@ -45,6 +45,8 @@ along with GCC; see the file COPYING3.  If not see
 #define LIB_SPEC GNU_USER_TARGET_LIB_SPEC " -latomic "
 #endif
 
+#define ICACHE_FLUSH_FUNC "__riscv_flush_icache"
+
 #define LINK_SPEC "\
 -melf" XLEN_SPEC "lriscv \
 %{shared} \
--- a/gcc/config/riscv/riscv.md
+++ b/gcc/config/riscv/riscv.md
@@ -1455,7 +1455,13 @@
    (match_operand 1 "pmode_register_operand")]
   ""
 {
+#ifdef ICACHE_FLUSH_FUNC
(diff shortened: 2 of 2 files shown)


palmer-dabbelt:
@jim-wilson: do you mind getting this upstream, and backported for 7.3.0?  It's the last ABI-visible fix.  I'll deal with the glibc side.

davidlt:
ping

I see this change landed in v2 for glibc and gcc 7.3.0 deadline is approaching.

palmer-dabbelt:
Jim is working on it now, I think it's pretty much ready to go.

jim-wilson:
If I add the patch to gcc now, then our linux gcc testing will start failing, as qemu does not emulate VDSOs.  It is about 80 extra gcc testsuite failures, which complicates my task of trying to reduce that number down to zero.  Otherwise, I'm OK with the patch and can merge it in anytime, er anytime that I can find the time to do it that is.

palmer-dabbelt:
Sorry, I forgot it wasn't all plumbed through yet.  I'll do that today.  When is the 7.3.0 deadline?

davidlt:
The last report was:
```
Status
======

GCC 7 is in regression and documentation fixes mode and it is time
to think about backports you want/need to do for GCC 7.3.  The plan
is to do a release candidate for GCC 7.3 in the second week of January
following by a release a week after that.

As usual the branch is in good shape for what is considered release
critical but you might want to eye status on non-primary/secondary
targets and languages.
```

palmer-dabbelt:
OK, we should be fine.

jim-wilson:
Patch upstream on mainline and gcc-7 branch.

PR #117, opened 2018-01-02
Implement naked attribute

kito-cheng:
gcc/ChangeLog
```
2018-01-02  Kito Cheng  <kito.cheng@gmail.com>

	* riscv/riscv-protos.h (riscv_output_return): New function.
	* config/riscv/riscv.c (machine_function): New naked_p field.
	  (riscv_attribute_table): Add table for target-specific attribute.
	  (riscv_output_return): Handle naked function.
	  (riscv_naked_function_p): Ditto.
	  (riscv_allocate_stack_slots_for_args): Ditto.
	  (riscv_warn_func_return): Ditto.
	  (riscv_compute_frame_info): Ditto.
	  (riscv_expand_prologue): Ditto.
	  (riscv_expand_epilogue): Ditto.
	  (riscv_set_current_function): Ditto.
	* config/riscv/riscv.md (simple_return): Call
	  riscv_output_return for handle naked function.
	* doc/extend.texi: Document naked attribute for RISC-V.
```


kito-cheng:
https://github.com/riscv/riscv-gnu-toolchain/issues/313

gcc/config/riscv/riscv.c
```diff
@@ -1809,6 +1826,17 @@ riscv_output_move (rtx dest, rtx src)
     }
   gcc_unreachable ();
 }
+
+
```
jim-wilson: Most functions are separated by a single blank line, and you have two here, but this is a minor detail.

gcc/config/riscv/riscv.c
```diff
@@ -3590,6 +3656,14 @@ riscv_expand_epilogue (bool sibcall_p)
   bool need_barrier_p = (get_frame_size ()
 			 + cfun->machine->frame.arg_pointer_offset) != 0;
 
+  if (cfun->machine->naked_p)
+    {
+      if (!sibcall_p)
```
jim-wilson: This is a little confusing at first glance, but since a naked function can't be a sibcall, this is just unconditionally calling emit_jump_insn.  It could maybe be written that way.

kito-cheng:
Yes, it's almost copied from another target, I've forbid sibcall for naked function now.

jim-wilson:
I'm wondering what the status of this is.  Kito, are you going to merge it into riscv-next?  Do you want it in gcc-8?  I don't believe it is appropriate for gcc-7, since it is a new feature.

palmer-dabbelt:
@jim-wilson We usually handle merging things ourselves, so if you think it's OK then you should merge it, and commit it upstream.  I agree it's not a gcc-7 target, but I think we should backport it to riscv-gcc-7.2.0.

kito-cheng:
@jim-wilson Yeah, I hope it can appear in gcc 8, thanks :)
@palmer-dabbelt I've cherry-picked to riscv-gcc-7.

btw, I plan to sync riscv-next with upstream again in next few days, it's OK for you guys?

jim-wilson:
On Tue, Jan 9, 2018 at 9:46 PM, Kito Cheng <notifications@github.com> wrote:

> @jim-wilson <https://github.com/jim-wilson> Yeah, I hope it can appear in
> gcc 8, thanks :)
> @palmer-dabbelt <https://github.com/palmer-dabbelt> I've cherry-picked to
> riscv-gcc-7.
>
> btw, I plan to sync riscv-next with upstream again in next few days, it's
> OK for you guys?
>

FYI, I was in the middle of cherry picking the -msave-restore patch.  But
it was my first attempt to cherry pick a patch onto the riscv-gcc-7.2.0
release branch, so I was being extra careful.  I tried running regressions,
saw some failures, got distracted by them, and didn't end up committing my
patch yet.  The patch you cherry picked is identical to mine.  I also see
you cherry picked the naked function patch.  Adding that to the FSF tree
for gcc-8 is my next task, though I don't plan to add it to the FSF gcc-7
branch.

I don't know what the rules are for the riscv-next branch, but I'm assuming
it can merge/rebase with the FSF top of tree at any time, so that won't
break anything I'm doing.

The regressions on  riscv-gcc-7.2.0 are for a g++ -shared test on *-elf
targets, which is not expected to work for newlib anyways.  This was broken
by the binutils shared library global pointer fix.  I have a patch in the
FSF binutils tree to disable shared library support for the elf target,
which would fix this, but might break other things if someone is trying to
use shared libraries with the elf/newlib targets.  Hopefully no one is
doing that, because it isn't supported, and isn't expected to work.

Jim

jim-wilson:
The patch has some bugs.  riscv_allocate_stack_slots_for_args and riscv_warn_func_return have the tests for a naked function reversed.  They both need a ! before the riscv_naked_function_p call.  I'm retesting after the bug fix to check for more problems.  The patch fixes will have to be backported into the riscv-gcc-7.2.0 branch.

The gcc tests should have been run before this patch was added to the riscv-gcc-7.2.0 branch.  You can run them for a single target by typing "make check" in a build tree.  I do this in a patched and unpatched tree and compare the results in the build-gcc-*-stage2/gcc/testsuite/{gcc,g++}/{gcc,g++}.sum files.  Or you can run them for all targets with a cd into the regression dir and "make".  This takes a while, even with a parallel make.  This compares against a white list and hence will catch new failures with just a single run in patched tree.

jim-wilson:
I've added the patch to FSF gcc-8, and push a fix for two bugs on the riscv-gcc-7.2.0 branch.

kito-cheng:
@jim-wilson Oh...It;s really my mistake during move those patch from our internal tree, I should always run testsuite before commit. I apology for my stupid behavior..

PR #118, opened 2018-01-23
[WIP]: Disable TLS copy relocs.

jim-wilson:
* config/riscv/riscv.c (riscv_legitimize_tls_address): Ifdef out
	code to force TLS_MODEL_LOCAL_EXEC when !flag_pic.

### DIFF
--- a/gcc/config/riscv/riscv.c
+++ b/gcc/config/riscv/riscv.c
@@ -1209,9 +1209,11 @@ riscv_legitimize_tls_address (rtx loc)
   rtx dest, tp, tmp;
   enum tls_model model = SYMBOL_REF_TLS_MODEL (loc);
 
+#if 0
   /* Since we support TLS copy relocs, non-PIC TLS accesses may all use LE.  */
   if (!flag_pic)
     model = TLS_MODEL_LOCAL_EXEC;
+#endif
 
   switch (model)
     {


jim-wilson:
Do not merge.  This is just for informational purposes.

The glibc testsuite tst-tls12 testcase fails because TLS copy relocs aren't working.  We can make it work by disabling a compiler optimization.

jim-wilson:
The real problem has finally been identified.  The linker is not allocating memory space for the tdata.dyn section.  This is a binutils bug, and an initial patch has been added, rendering this moot.

richfelker:
Please disable this feature. It doubles memory usage **per thread** of each TLS object affected by it and has no benefit.

richfelker:
And as I just noted in https://sourceware.org/bugzilla/show_bug.cgi?id=23825, it breaks the ability to have extensible structs/arrays in shared library TLS (because copy relocations always break this).

jim-wilson:
This is being discussed in
   https://github.com/riscv/riscv-elf-psabi-doc/issues/122
which is the proper place to discuss RISC-V ABI issues.

MaskRay:
Just a reminder. @jim-wilson Will you send a patch to upstream GCC?

jim-wilson:
I'm testing a patch now.

jim-wilson:
I committed the patch upstream.

PR #119, opened 2018-01-24
DON'T MERGE: Set invalid flag for sNaN inputs to isless etc.

aswaterman:


aswaterman:
This isn't required by C99/C11, but it is expected by the glibc test suite.  We are still determining the best solution.

jim-wilson:
I have a revised patch that adds an expander, and then has two matching patterns, the current one uses ! HONOR_SNAN and the new slower one uses HONOR_SNAN.  There are also a few indentation changes to follow the default formatting rules.  This works for a testcase extracted from glibc.  I'm running a full native riscv64-linux gcc and glibc build and test now.

[isless-snan.patch.txt](https://github.com/riscv/riscv-gcc/files/2448072/isless-snan.patch.txt)

aswaterman:
Thanks, Jim - that looks like the way to go.  I will close this PR.

PR #123, opened 2018-02-23
-pthread and -fopenmp support for riscv-unknown-elf

rishikhan:
Added -pthread command line support for riscv-*-elf. Instructed libgomp that riscv-unknown-elf does not have semaphore support.

This gives you pthread and openmp support assuming you have a working libpthread.a

### DIFF
--- a/config.sub
+++ b/config.sub
@@ -1396,7 +1396,7 @@ case $os in
 	      | -ekkobsd* | -kfreebsd* | -freebsd* | -riscix* | -lynxos* \
 	      | -bosx* | -nextstep* | -cxux* | -aout* | -elf* | -oabi* \
 	      | -ptx* | -coff* | -ecoff* | -winnt* | -domain* | -vsta* \
-	      | -udi* | -eabi* | -lites* | -ieee* | -go32* | -aux* \
+	      | -udi* | -eabi* | -lites* | -ieee* | -go32* | -aux* | -hcos* \
 	      | -chorusos* | -chorusrdb* | -cegcc* | -glidix* \
 	      | -cygwin* | -msys* | -pe* | -psos* | -moss* | -proelf* | -rtems* \
 	      | -midipix* | -mingw32* | -mingw64* | -linux-gnu* | -linux-android* \
--- a/gcc/config.gcc
+++ b/gcc/config.gcc
@@ -2033,6 +2033,7 @@ riscv*-*-linux*)
 	;;
 riscv*-*-elf* | riscv*-*-rtems*)
 	tm_file="elfos.h newlib-stdint.h ${tm_file} riscv/elf.h"
+	extra_options="$extra_options gnu-user.opt"
 	case "x${enable_multilib}" in
 	xno) ;;
 	xyes) tmake_file="${tmake_file} riscv/t-elf-multilib" ;;
--- a/gcc/config.gcc
+++ b/gcc/config.gcc
@@ -2031,7 +2031,7 @@ riscv*-*-linux*)
 	# automatically detect that GAS supports it, yet we require it.
 	gcc_cv_initfini_array=yes
 	;;
-riscv*-*-elf* | riscv*-*-rtems* | riscv*-*-elf_hcos)
+riscv*-*-elf* | riscv*-*-rtems* | riscv*-*-hcos*)
 	tm_file="elfos.h newlib-stdint.h ${tm_file} riscv/elf.h"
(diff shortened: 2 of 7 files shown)


jim-wilson:
As I mentioned with your newlib pthread patches, this stuff can only work with an OS, and riscv*-*-elf is not an OS.  You need to give your OS a name, and then add configure triplets for that OS name.

rishikhan:
Is the issue this line [and the corresponding .ac file]:
riscv-unknown-elf*)

or something else?

pthreads does not need an OS. This is only the case if it is preemptive. Each core has a core with a pthread on it. They wait for a pthread_create message from the first core, look up their address [in a predefined array] and jump to that address. For all of the locking and barrier semantics, the threads simply halt or spin until the trigger is met.

 I guess you can say we have an OS, but all it does is shuttle syscalls to the host.

In any case, the typical triplet is cpu-vendor-os. Are you suggesting I replace elf with a name? If so, I can do that.

rishikhan:
Jim,
   I want everything to compile 'as if' it were elf. I tried adding an OS and ended up that every riscv-gnu-toolchain project's config.sub needs to be modified with that new OS. So we're calling our stub hcos. Do you think it's fine to make the OS=elf_hcos so that all of the different subcomponents work as if it were elf?

jim-wilson:
I'm concerned about breaking riscv*-elf support for other people.  You are adding stuff that only works in your environment, which is not the same as the UCB environment, or the SiFive environment, or Liviu's environment, etc.  At some point these things need to be distinguished somehow.

Also, it is not normal for a *-*-elf target to have pthread support, libgomp, and other related things.  There are only a few exceptions, and they are all targets which only appear in one hardware/software environment, like the processor used in the Playstation 3.

There is precedent for merging OS names into ELF.  We used to have an arm-*-ecos-elf target, and we still have an arm-*-symbianelf target.  There is also precedent for distinguishing between libraries with stuff at the end, so we have linux-gnu, linux-uclibc, linux-musl.  Underscores are not common in triplets, but there are some obvious examples like x86_64 and aarch64_be so this is OK.

Another point here is that you are submitting this against the gcc-7.2.0 branch which is already obsolete, as there was a gcc-7.3.0 release, we just haven't gotten around to updating this tree yet.  A patch like this would likely not be accepted in the upstream FSF gcc-7.3.0 tree because it is a new feature.  It could still get into gcc-8 if submitted before the release branch is made.  At some point you might want to start submitting patches upstream instead of into the github riscv projects.

rishikhan:
I appreciate the comments. I'll work on fixing it. The things I'm adding are solely for hcos in gcc [and newlib], so they don't really do anything to elf (honestly, it's a few header files and config params ... and pretty much 0 code). I see your comment on 'extra_options' and agree. Further, if you want, I can remove the -pthread definition and add it to elf_hcos.h. Then there will be no impact on elf. Agree?

I do have concern about building out an OS name that is not an extension of elf because, honestly, it's doing nothing different than elf in any of these codes [except header and config changes] and all of the 'code' is in the pthread library. If I make a new OS name, then every single other project needs to be modified (binutils, dejagnu, glibc, qemu, etc) to accept that new OS, or they won't compile. When in reality compiling them for 'elf' is perfectly fine.

rishikhan:
> The elf_hcos name should be reviewed and approved upstream before you standardize on it. There is some risk that other upstream maintainers might not like it and might ask for changes to the name.

Which maintainers should I ask? it seems that ALL of the repos would have to agree (gcc, newlib, binutils, etc). Correct?

jim-wilson:
Once one major package agrees to accept the new triple it is unlikely others would object.  The central point for this would be the config.sub maintainers which is a separate package which you may have to deal with eventually, but contracting the gcc maintainers would cover a lot of this.  I think most binutils maintainers are probably also reading or at least skimming gcc mailing lists for instance.

Yes, the gcc changes here aren't very significant.  My main concern is what happens with the newlib patches.  I don't think that adding pthread support to riscv-elf is right.  If you need a new configure triple for newlib, then you may as well use it for your gcc patches too.

rishikhan:
I moved this to a OS name. Is this good? [if so, I'll send the patch to binutils and newlib].

jim-wilson:
Adding patches to gcc-7.2 isn't useful.  I think gcc patches should go upstream.

kito-cheng:
As Jim's suggestion, this patch should go upstream directly, so close this PR.

PR #129, opened 2018-04-25
Don't depends on __NR_rt_sigreturn as as signature

guoren83:
__NR_rt_sigreturn is from linux unistd.h and it's different in history linux kernel version. So depends on the __NR_rt_sigreturn number is bad smell.
So you can define your own magic number as a signature to detect the rt_sigframe.

### DIFF
--- a/libgcc/config/riscv/linux-unwind.h
+++ b/libgcc/config/riscv/linux-unwind.h
@@ -25,7 +25,8 @@
 #include <stdint.h>
 #include <sys/ucontext.h>
 
-#define LI_A7_8B 0x08b00893
+#define LI_A7_55 0x05500893
+#define LI_A7_AA 0x0aa00893
 #define ECALL    0x00000073
 
 #define MD_FALLBACK_FRAME_STATE_FOR riscv_fallback_frame_state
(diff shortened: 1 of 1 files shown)


guoren83:
It 's a RFC, just for discussion.

aswaterman:
This is already part of the ABI so can't be easily changed.

Regardless, several other ports (x86, alpha, mips, ppc, sparc, etc.) do it the same way we're doing it, so I don't think we're doing anything too weird.

palmer-dabbelt:
I agree with Andrew: we can't change this as it would make GCC incompatible with glibc.  Are you sure __NR_rt_sigreturn can change values?  I was under the impression that's already part of the stable ABI on Linux, but if not then we'll need to enforce it ourselves.

jim-wilson:
Using __NR_rt_sigreturn is the standard way to detect a signal call frame.  I count 11 targets mentioning this in comments, and the total number is more, because not all of them have a comment, e.g. RISC-V doesn't have one.

Changing __NR_rt_sigreturn is a major ABI break.  If that number changes, then likely everything else is broken too, and we need a new toolchain anyways.

Different targets can have different values for __NT_rt_sigreturn, but that isn't relevant.  Only the RISC-V value matters here.

It is possible that some old versions had unstable numbers for __NR_rt_sigreturn before it was accepted as a standard syscall, but that isn't relevant either, since only the current RISC-V value matters.

Your proposed solution is unsafe, because the code sequence could potentially occur in user code, for some badly optimized code for instance, or some badly written asms, etc.

I don't see any problem here that needs fixing.

guoren83:
Thx aswaterman palmer & jim's explain, I got them.
Seems __NR_rt_sigreturn won't change in linux.

PR #134, opened 2018-07-03
Back port patches to riscv-gcc-8.1.0 branch from trunk

kito-cheng:
Tested with rv64gc, rv32gc, rv64imafc, rv32imafc, rv64imac and rv32imac.


jim-wilson (approved):
There are a couple of optimization patches that we don't really need, but they shouldn't break anything so should be OK.

jim-wilson:
It probably makes sense to update riscv-gnu-toolchain to point at the new riscv-gcc-8.1.0 top of tree.

PR #136, opened 2018-07-18
Backport rv32e support to riscv-gcc-8.1.0

palmer-dabbelt:
I think this is the only patch we need.  I haven't even tried compiling this yet...


kito-cheng:
I’ve merged that before, this PR seems like only modify the changelog fils?

kito-cheng:
I think the missing part is binutils, do you have plan to upgrade/merge that? I minght have time next week if you have no time to do.

PR #151, opened 2019-02-24
Riscv gcc 8.3.0

galak:
I rebased the riscv-gcc-8.2.0 branch on the `gcc-8_3_0-release` release tag.  Wanted to share that work.


jim-wilson:
Have you tried a regression test?  We should keep this working.  cd into the regression dir,  type make, and it will build about 26 targets, run the gcc testsuite, exclude known failures in white lists, and report an error if there are new failures.  If the new failures are benign, we can add them to the white lists.  Otherwise, they need to be fixed.  This is best done on a machine with multiple cores using make -jX to use as many as feasible.  This command can be restarted multiple times if necessary, as long as you don't touch the source tree, as that may force everything to rebuild from scratch.

jim-wilson:
The branch looks correct.  There is one new upstream regression.  The testcase g++.dg/pr83239.C is failing for most if not all targets.  This is reported upstream as FSF GCC bug 86153.

This was submitted as a git rebase against the riscv-gcc-8.2.0 branch which is not OK.  I created a new riscv-gcc-8.3.0 branch and pushed your work there.  I'm also working on a riscv-gnu-toolchain patch to use the new riscv-gcc branch.

PR #162, opened 2019-07-30
Raise error on unexpected ISA string at end

mablinov:
Hi all, this patch backports the error handling code from binutils-gdb.

Raise an error when the user inputs an unexpected arch string past the expected end, rather than silently pruning the arch string.

### DIFF
--- a/gcc/common/config/riscv/riscv-common.c
+++ b/gcc/common/config/riscv/riscv-common.c
@@ -513,6 +513,13 @@ riscv_subset_list::parse (const char *arch, location_t loc)
   if (p == NULL)
     goto fail;
 
+  if (*p != '\0')
+    {
+      error_at (loc, "-march=%s: unexpected ISA string at end: %<%s%>",
+               arch, p);
+      return NULL;
+    }
+
   return subset_list;
 
 fail:
--- a/gcc/common/config/riscv/riscv-common.c
+++ b/gcc/common/config/riscv/riscv-common.c
@@ -517,7 +517,7 @@ riscv_subset_list::parse (const char *arch, location_t loc)
     {
       error_at (loc, "-march=%s: unexpected ISA string at end: %<%s%>",
                arch, p);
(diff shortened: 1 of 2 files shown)


kito-cheng:
Hi Maxim, could you write a test case for that? thanks :)

gcc/common/config/riscv/riscv-common.c
```diff
@@ -513,6 +513,13 @@ riscv_subset_list::parse (const char *arch, location_t loc)
   if (p == NULL)
     goto fail;
 
+  if (*p != '\0')
+    {
+      error_at (loc, "-march=%s: unexpected ISA string at end: %<%s%>",
+               arch, p);
+      return NULL;
```
kito-cheng: `goto fail;` instead of `return NULL;` here, because subset_list need to free

mablinov:
BTW, is it ok to force push stuff like I did above? I wanted to squash a commit that just removed some newlines, and make the history look a bit cleaner. Should I just leave even the small formatting-issue commits as they are?

jim-wilson:
A patch needs to be upstream before you can backport it.  I don't see this patch in FSF GCC mainline.  If we start adding local patches that aren't upstream we are likely to lose track of them.

It is common practice to force push pull requests.  This gives a cleaner git log once it gets merged in.  Rebasing is OK too, if someone else adds a conflicting patch.

mablinov:
Ok, I've submitted it to gcc-patches@gcc.gnu.org

mablinov:
Patch merged upstream, closing.

kito-cheng:
@mablinov thanks!

I've back-port to `riscv-gcc-8.3.0` branch, but I would bump it into riscv-gnu-toolchian later, because I've some other patch need bump together.
https://github.com/riscv/riscv-gcc/commit/a3fb5274fbd3858d9dfa78490a9bbdc47dd6ce73

PR #165, opened 2019-08-12
Include rvintrin.h bitmanip intrinsics header file

mablinov:
Hi all,

This patch simply adds the `rvintrin.h` intrinsics header file into the build system.

Destination directory:
`${prefix}/lib/gcc/riscv32-unknown-elf/10.0.0/include/rvintrin.h`
Tested, GCC finds it without issues.

### DIFF
--- a/gcc/config.gcc
+++ b/gcc/config.gcc
@@ -520,6 +520,7 @@ pru-*-*)
 	;;
 riscv*)
 	cpu_type=riscv
+	extra_headers="rvintrin.h"
 	extra_objs="riscv-builtins.o riscv-c.o"
 	d_target_objs="riscv-d.o"
 	;;
--- a/gcc/config/riscv/rvintrin.h
+++ b/gcc/config/riscv/rvintrin.h
@@ -1,0 +1,699 @@
+/*
+ *  RISC-V "B" extension proposal intrinsics and emulation
+ *
+ *  Copyright (C) 2019  Clifford Wolf <clifford@clifford.at>
+ *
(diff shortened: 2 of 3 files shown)


aswaterman:
Is this upstream yet? (Same rationale that Jim wrote on https://github.com/riscv/riscv-gcc/pull/162 applies).

jim-wilson:
This is a pull request for the bitmanip branch.  Maintaining patches in a git branch is much nicer than maintaining them as patch files in riscv-bitmanip.  I'm OK with having them here, as long as people understand that the review for this tree does not eliminate the need for the upstream review, and that further changes may be requested by the upstream review, as I'm not doing full FSF style patch reviews here.

I'm assuming that Clifford Wolf is maintaining this tree, but I can help if necessary.  Or we can maybe add Max to the list of people with write access since I expect he will be doing more bitmanip toolchain work.

jim-wilson:
Not directly related to this, but the testcases in riscv-bitmanip could be converted into binutils and gcc testcases too.  This would be gas/testsuite/gas/riscv for binutils and gcc/testsuite/gcc.target/riscv for gcc.

aswaterman:
Oops, missed that, thanks Jim. Maintaining a branch here seems preferable to maintaining patch files, and we have already gone in this direction for Spike in this github org.

cliffordwolf:
Thanks! I've cherry-picked b4f0e10 from your branch so I don't re-add my hotfix in
a63392d that I just removed via rebasing the riscv-bitmanip branch.

PR #166, opened 2019-08-13
Add canonical parsing of zbX -march flags

mablinov:
Hi all,

This patch adds "-march=rv32imb_[zb*]*" type parsing. It mirrors my pull request in riscv-binutils-gdb/riscv-bitmanip in terms of functionality.

I have classified each instruction, and the "zbx" selectively enables the corresponding groups. The 'b' flag still exists, though. I had a good reason for leaving it in, but I don't remember it anymore... 

All flags are additive.

There are doubtless some bits I forgot to update or re-classify.

### DIFF
--- a/gcc/common/config/riscv/riscv-common.c
+++ b/gcc/common/config/riscv/riscv-common.c
@@ -75,17 +75,13 @@ typedef struct riscv_parse_config
 
   /* Lower-case prefix string for error printing
      and internal parser usage, e.g. "sx", "z".  */
-  const char *prefix_lower;
-
-  /* Upper-case prefix string for error printing,
-     e.g. "SX". Used as "Error: SX extension ..."  */
-  const char *prefix_upper;
+  const char *prefix;
 
   /* Predicate which is used for checking whether
      this is a "known" extension. For 'x' and 'sx',
      it always returns true (since they are by
      definition non-standard and cannot be known.  */
-  int (*ext_valid_p) (const char *);
+  bool (*ext_valid_p) (const char *);
 } riscv_parse_config_t;
 
 /* Subset info.  */
@@ -506,7 +502,7 @@ riscv_subset_list::parse_prefixed_ext (const char *p,
       if (!config->ext_valid_p (subset))
 	{
 	  error_at (m_loc, "%<-march=%s%>: Invalid or unknown %s ISA extension: %qs",
-		    m_arch, config->prefix_upper, subset);
+		    m_arch, config->prefix, subset);
 	  free (subset);
 	  return NULL;
 	}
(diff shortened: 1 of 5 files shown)


gcc/common/config/riscv/riscv-common.c
```diff
@@ -441,29 +489,215 @@ riscv_subset_list::parse_sv_or_non_std_ext (const char *p,
...
+const char * const riscv_std_s_ext_strtab[RISCV_STD_S_EXT_COUNT] =
+  {
+
+  };
+
+/* Find the index corresponding to the z-extension
+   name `ext'. If not found, return -1.  */
+
+int
+riscv_std_z_ext_index (const char *ext)
```
kito-cheng: Seems like this function no longer used

gcc/common/config/riscv/riscv-common.c
```diff
@@ -441,29 +489,215 @@ riscv_subset_list::parse_sv_or_non_std_ext (const char *p,
...
+    }
+
+  return -1;
+}
+
+/* Predicator function for x-prefixed extensions.
+   Anything goes, except the literal 'x'.  */
+
+static int
+riscv_ext_x_valid_p (const char *arg)
```
kito-cheng: `riscv_ext_*_valid_p` could be removed or simplified once we have ext_info_t table.

gcc/common/config/riscv/riscv-common.c
```diff
@@ -33,6 +33,61 @@ along with GCC; see the file COPYING3.  If not see
...
+  /* Class of the extension. */
+  riscv_isa_ext_class_t ext_class;
+
+  /* Lower-case prefix string for error printing
+     and internal parser usage, e.g. "sx", "z".  */
+  const char *prefix_lower;
+
+  /* Upper-case prefix string for error printing,
+     e.g. "SX". Used as "Error: SX extension ..."  */
+  const char *prefix_upper;
```
kito-cheng: Like review comment in binutils side, just merge `prefix_lower` and `prefix_upper` is fine.

gcc/common/config/riscv/riscv-common.c
```diff
@@ -576,8 +808,26 @@ riscv_parse_arch_string (const char *isa, int *flags, location_t loc)
...
+      *bitmanip_flags |= OPTION_MASK_ZBM;
+      *bitmanip_flags |= OPTION_MASK_ZBP;
+      *bitmanip_flags |= OPTION_MASK_ZBR;
+      *bitmanip_flags |= OPTION_MASK_ZBS;
+      *bitmanip_flags &= ~OPTION_MASK_ZBT;
+    }
+
+  /* Handle specific bitmanip ISA extensions.  */
+  if (riscv_have_bitmanip_z_ext_string_p (subset_list))
+    riscv_set_bitmanip_z_ext_flags (opts, subset_list);
```
kito-cheng: This could be simplified as iterate ext_info_t table.

gcc/common/config/riscv/riscv-common.c
```diff
@@ -441,29 +489,215 @@ riscv_subset_list::parse_sv_or_non_std_ext (const char *p,
...
     }
 
   return p;
 }
 
+#define RISCV_STD_Z_EXT_COUNT 9
+
+const char * const riscv_std_z_ext_strtab[RISCV_STD_Z_EXT_COUNT] =
+  {
+   "zbb", "zbc", "zbe", "zbf", "zbm", "zbp", "zbr", "zbs", "zbt"
```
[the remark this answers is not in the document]
kito-cheng: Oh, I forgot gcc is C++ now, so it could use pointer to member data, that's would be more type-safe.

kito-cheng:
It's riscv-bitmanip branch which is not used by default branch of riscv-gnu-toolchain, so I am fine to merge that if @cliffordwolf is OK.

And seems there is a conflict need to resolve before merge.

kito-cheng:
We've handle that now.

PR #167, opened 2019-08-27
Add intrinsics function for rolw

kito-cheng:
Just a proof of concept, I didn't run the testsuite or any serious testing.
```
$ cat x.c
#include <rvintrin.h>

int rolw(int a, int b){
  return _rv32_rol(a, b);
}
int rolw_bi(int a, int b){
  return __builtin_riscv_rolw(a, b);
}

$ riscv64-unknown-elf-gcc x.c -S -o - -O3 -fdump-tree-all -march=rv64ib -mabi=lp64
...
rolw:
 #APP
# 137 "/scratch/kitoc/riscv-gnu-workspace/riscv-gnu-toolchain-b-ext/build/install/lib/gcc/riscv64-unknown-elf/10.0.0/include/rvintrin.h" 1
        rolw    a0, a0, a1
# 0 "" 2
 #NO_APP
        sext.w  a0,a0
        ret
...
rolw_bi:
        rolw    a0,a0,a1
        ret
...
```

### DIFF
--- a/gcc/config/riscv/bitmanip.md
+++ b/gcc/config/riscv/bitmanip.md
@@ -260,6 +260,16 @@
   "ror%i2\t%0,%1,%2"
   [(set_attr "type" "bitmanip")])
 
+(define_expand "riscv_rolw"
+  [(match_operand:SI 0 "register_operand" "=r")
+   (match_operand:SI 1 "register_operand" "r")
+   (match_operand:SI 2 "register_operand" "r")]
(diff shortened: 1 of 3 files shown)


kito-cheng:
@cliffordwolf

cliffordwolf:
Looks good to me. (By that I mean it appears to do what I want. I'm not really qualified to review any GCC patches that touch any `.md` files. ;)

PR #173, opened 2019-12-27
Enable the constraints checking for assembler by default if possible.

Nelson1225:
I get the request recently that the assembler's constraints checking break hardware exception test cases.  Therefore I have disabled the checking in assembler by default, and people don't get the constraints error when writing assembly code.  But we should enable the checking for the code which the compiler emits.

Beside, GCC also checks whether the assembler supports the options used to enable and disable the constraints checking.  The basic logic is shown as follows,

| | GAS support constraints options | GAS doesn't support constraints options
--- | --- | ---
GCC default | Output `.option checkconstraints` | Don't output the `.option checkconstraints`
GCC `-mcheck-constraints` | Output `.option checkconstraints` | Report error message that GAS doesn't support `.option checkconstraints`
GCC `-mno-check-constraints` | Don't output the`.option checkconstraints`| Don't output the`.option checkconstraints`

One more thing is that, the `gcc_GAS_CHECK_FEATURE` doesn't seem to work for checking the `.option xxx`.  I think the reason may be that GAS report warning message rather than error when the `.option xxx` isn't supported.  However, I have added both GAS options and .option directives for the
constraints checking, so I check them both in the `gcc_GAS_CHECK_FEATURE`.

	gcc/
	* config.in: Regenerated.
	* configure: Likewise.
	* configure.ac: Check whether GAS support constraints options.
	* config/riscv/riscv.c (riscv_option_override): If user don't set the
	`m[no-]check-constraints` option and GAS support the constraints
	checking, then GCC should enable it by default. If user set the
	`mcheck-constraints` options but the checking isn't supported in GAS,
	then we should report error message.
	(riscv_file_start): Emit `.option checkconstraints` if needed.
	* config/riscv/riscv.opt (mcheck-constraints): New option.

	gcc/testsuite/
	* gcc.target/riscv/constraints-1.c: New.
	* gcc.target/riscv/constraints-2.c: Likewise.

### DIFF
--- a/gcc/config/riscv/riscv.c
+++ b/gcc/config/riscv/riscv.c
@@ -4399,6 +4399,9 @@ riscv_file_start (void)
   if (! riscv_mrelax)
     fprintf (asm_out_file, "\t.option norelax\n");
 
+  if (riscv_mcheck_constraints)
+    fprintf (asm_out_file, "\t.option checkconstraints\n");
+
   if (riscv_emit_attribute_p)
     riscv_emit_attribute ();
 }
@@ -4582,6 +4585,21 @@ riscv_option_override (void)
   if (riscv_emit_attribute_p)
     error ("%<-mriscv-attribute%> RISC-V ELF attribute requires GNU as 2.32"
 	   " [%<-mriscv-attribute%>]");
+#endif
+
+  /* If user don't set the `m[no-]check-constraints` option and assembler
+     support the constraints checking, then GCC should enable it by default.
+     If user set the `mcheck-constraints` options but the checking isn't
+     supported in assembler, then we should report error here.  */
+  if (riscv_mcheck_constraints < 0)
+#ifdef HAVE_AS_RISCV_CHECK_CONSTRAINTS
+    riscv_mcheck_constraints = 1;
+#else
+    riscv_mcheck_constraints = 0;
+
+  if (riscv_mcheck_constraints)
+    error ("%<-mcheck-constraints%> RISC-V constraints checking is unsupported"
+	   " in GNU as [%<-mcheck-constraints%>]");
 #endif
 }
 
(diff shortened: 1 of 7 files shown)


Nelson1225:
The related support in binutils https://github.com/riscv/riscv-binutils-gdb/pull/193

gcc/config/riscv/riscv.c
```diff
@@ -4583,6 +4586,21 @@ riscv_option_override (void)
...
+     supported in assembler, then we should report error here.  */
+  if (riscv_mcheck_constraints < 0)
+#ifdef HAVE_AS_RISCV_CHECK_CONSTRAINTS
+    riscv_mcheck_constraints = 1;
+#else
+    riscv_mcheck_constraints = 0;
+
+  if (riscv_mcheck_constraints)
+    error ("%<-mcheck-constraints%> RISC-V constraints checking requires GNU"
+	   " as 2.33.1 [%<-mcheck-constraints%>]");
```
jim-wilson: Giving the gas version number in the error message looks a little odd, as this feature isn't in upstream GNU as, upstream GNU as is already to 2.34, and is only in vector branch with the riscv-binutils-gdb  #193 pr which doesn't change the gnu as version number.  I think it is better without version number until the patches are upstream and there is a proper binutils release we can refer to.

Nelson1225: OK, I update the message as follows,
"RISC-V constraints checking is unsupported in GNU as"

Nelson1225:
> You can pass --fatal-warnings to GNU as to change warnings into errors. That might fix your configure patch problem.

Thanks this works :)

kito-cheng:
Verified on my side, I gonna merge this.

Nelson1225:
Thanks for helps :)

PR #177, opened 2020-02-26
Fix rv64 check for rotldi3

bdwyatt:
On an RV32 target I was seeing a single rol instruction being emitted for code that was rotating a uint64_t, and tracked it down to the missing RV64 check on the rotldi3 description.

### DIFF
--- a/gcc/config/riscv/bitmanip.md
+++ b/gcc/config/riscv/bitmanip.md
@@ -290,7 +290,7 @@
   [(set (match_operand:DI 0 "register_operand" "=r")
 	(sign_extend:DI (rotate:SI (match_operand:SI 1 "register_operand" "r")
 				   (match_operand:SI 2 "register_operand" "r"))))]
-  "TARGET_BITMANIP"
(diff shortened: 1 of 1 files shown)


jim-wilson:
Do you have an FSF copyright assignment?  FSF copyright assignments are required to contribute to FSF code.  If we don't require FSF copyright assignments here, then we won't be able contribute this work to the FSF when it is done.  This change is simple enough and obvious enough that we can accept it without an assignment, but if you plan to do more work, then you should either have a verified assignment or you should just submit bugs and let someone with an assignment write the patch to fix it.

The patch looks correct but incomplete to me.  The immediately following rotlsi3_sext needs the same fix.

bdwyatt:
I do not currently have an FSF copyright assignment, but it is something I can look into.

Until I can resolve that, I will stick to reporting bugs.

cliffordwolf:
> This change is simple enough and obvious enough that we can accept it without an assignment

:+1:

PR #185, opened 2020-05-29
Refine use of sbseti for loading constants.

jim-wilson:
This is undesirable for bits 0 to 31 which should be loaded with addi or lui.
It fails for bit 32 which must be signed extended when it is a 32-bit value.
This case leads to an internal compiler error.  It can be used for a 64-bit
value with bit 32 set, but this can't be easily implemented.  So only allow it
for bits in the upper 32-bits of a 64-bit value.


jim-wilson:
No comments.  I'll self approve and commit.

PR #188, opened 2020-06-10
RISC-V: Make __divdi3 handle div by zero same as hardware.

virginiemoser:
The ISA manual specifies that divide by zero always returns -1 as the result.
We were failing to do that when the dividend was negative.

Original patch from Virginie Moser.

	libgcc/
	* config/riscv/div.S (__divdi3): For negative arguments, change bgez
	to bgtz.

### DIFF
--- a/libgcc/config/riscv/div.S
+++ b/libgcc/config/riscv/div.S
@@ -107,10 +107,12 @@ FUNC_END (__umoddi3)
   /* Handle negative arguments to __divdi3.  */
 .L10:
   neg   a0, a0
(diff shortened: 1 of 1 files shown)


kito-cheng:
riscv-gnu-toolchain is moving forward to gcc 10 now, so I would prefer back port that into 10.1 instead of 9.2.

kito-cheng:
But I am fine to backport that to 9.2, @jim-wilson any comment on this? I'll merge this if you don't have strong opinion on this.

PR #189, opened 2020-06-11
RISC-V: Make __divdi3 handle div by zero same as hardware.

virginiemoser:
The ISA manual specifies that divide by zero always returns -1 as the result.
We were failing to do that when the dividend was negative.

Original patch from Virginie Moser.

	libgcc/
	* config/riscv/div.S (__divdi3): For negative arguments, change bgez
	to bgtz.


kito-cheng:
I've backport several patches to `riscv-gcc-10.1.0` branch from trunk included this one, thanks :)

PR #192, opened 2020-06-28
Fix compilation error for riscv_remove_unneeded_save_restore_calls

fakepaper56:
Hi all,

This patch updates sp when removing save-restore. 
Gcc has compilation error when compiled the c file with option -msave-restore.

$ cat sample.c
int foo(int);
int bar(int a, int b, int c, int d, int e, int f, int g, int h, int u) {
  return foo(u);
}

$ riscv32-unknown-elf-gcc -march=rv32ima -mabi=ilp32 -S -Os -msave-restore sample.c
$ cat sample.s
bar:
    lw  a0,16(sp)
    tail    foo

`lw  a0,16(sp)` using wrong offset, the offset should be zero.

### DIFF
--- a/gcc/config/riscv/riscv-sr.c
+++ b/gcc/config/riscv/riscv-sr.c
@@ -171,6 +171,42 @@ riscv_sr_match_epilogue (void)
   return start;
 }
 
+/* If the function has stack incomming arguments, find all the uses of sp
+   between its prologue and its epilogue and update their offset  */
+static void
+update_sp_use (rtx_insn *prologue, rtx_insn *epilogue)
+{
+  rtx_insn *insn;
+
+  /* Skip if the function has no stack incomming arguments */
+  if (crtl->args.size == 0)
+    return;
+
+  for (insn = NEXT_INSN (prologue); insn != epilogue; insn = NEXT_INSN (insn))
+    if (INSN_P (insn) && GET_CODE (PATTERN (insn)) == SET)
+      {
+	rtx pat_src = SET_SRC (PATTERN (insn));
+
+	/* case assigning sp value */
+	if (GET_CODE (pat_src) == PLUS && REG_P (XEXP (pat_src, 0))
+	    && REGNO (XEXP (pat_src, 0)) == STACK_POINTER_REGNUM)
+	  {
+	    INTVAL (XEXP (pat_src, 1)) -= TARGET_RVE ? 12 : 16;
+	    continue;
+	  }
+
+	/* case loading stack incoming arguments  */
+	if (GET_CODE (pat_src) == SIGN_EXTEND
+	    || GET_CODE (pat_src) == ZERO_EXTEND)
+	  pat_src = XEXP (pat_src, 0);
+
+	if (MEM_P (pat_src) && GET_CODE (XEXP (pat_src, 0)) == PLUS
+	    && REG_P (XEXP (XEXP (pat_src, 0), 0))
+	    && REGNO (XEXP (XEXP (pat_src, 0), 0)) == STACK_POINTER_REGNUM)
+	  INTVAL (XEXP (XEXP (pat_src, 0), 1)) -= TARGET_RVE ? 12 : 16;
+      }
+}
+
 /* Helper for riscv_remove_unneeded_save_restore_calls.  If we match the
    prologue instructions but not the epilogue then we might have the case
    where the epilogue has been optimized out due to a call to a no-return
@@ -218,6 +254,7 @@ check_for_no_return_call (rtx_insn *prologue)
     fprintf (dump_file,
 	     "Prologue call to riscv_save_0 followed by noreturn call, "
 	     "removing prologue.\n");
+  update_sp_use (prologue, NULL);
   remove_insn (prologue);
 }
 
@@ -455,8 +492,9 @@ riscv_remove_unneeded_save_restore_calls (void)
   REG_NOTES (insn) = REG_NOTES (call);
   SIBLING_CALL_P (insn) = 1;
 
-  /* Now update the prologue and epilogue to take account of the
+  /* Now update the prologue, epilogue and use of sp to take account of the
      changes within the function body.  */
+  update_sp_use (prologue_matched, epilogue_matched);
   remove_insn (prologue_matched);
   remove_insn (NEXT_INSN (NEXT_INSN (NEXT_INSN (epilogue_matched))));
   remove_insn (NEXT_INSN (NEXT_INSN (epilogue_matched)));
--- a/gcc/testsuite/gcc.target/riscv/save-restore-10.c
+++ b/gcc/testsuite/gcc.target/riscv/save-restore-10.c
@@ -1,0 +1,20 @@
+/* { dg-options "-Os -msave-restore" } */
+
+/* With -msave-restore, gcc should keep correct sp offset for assignment when
+   removing redundant save-restore calls */
+
+extern int
+foo (int);
+
+int
+#if defined(__riscv_32e)
+bar (int a, int b, int c, int d, int e, int f, int u)
+#else
+bar (int a, int b, int c, int d, int e, int f, int g, int h, int u)
+#endif
+{
+  int *p = &u;
+  return foo (p[a]);
+}
+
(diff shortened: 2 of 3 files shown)


jim-wilson:
Patches must be sent upstream.  And upstream requires copyright assignments, unless the patch is small enough and trivial enough.  But it is OK to file bug reports without patches here.

kito-cheng:
@fakepaper56 
You need add a testcase to gcc testsuite.

@jim-wilson 
I confirmed with @kuanlinchentw, @fakepaper56 is employee of Andes, so he should not has copyright assignment issue, but I think this should go upstream directly.

jim-wilson:
The patch looks reasonable, but it won't fix every possible case of this problem.  Consider this testcase using a slightly older compiler to show the assembly, it also crashes with current compiler.

rohan:2064$ cat tmp.c
int foo(int);
int bar(int a, int b, int c, int d, int e, int f, int g, int h, int u) {
  char *ptr = (char *)&u;
return foo(ptr[a]);
}
rohan:2065$ riscv64-unknown-elf-gcc -O1 -msave-restore -S tmp.c
rohan:2066$ cat tmp.s
	.file	"tmp.c"
	.option nopic
	.attribute arch, "rv64i2p0_m2p0_a2p0_f2p0_d2p0_c2p0"
	.attribute unaligned_access, 0
	.attribute stack_align, 16
	.text
	.align	1
	.globl	bar
	.type	bar, @function
bar:
	call	t0,__riscv_save_0
	addi	a5,sp,16
	add	a0,a5,a0
	lbu	a0,0(a0)
	call	foo
	tail	__riscv_restore_0
	.size	bar, .-bar
	.ident	"GCC: (GNU) 9.2.0"
rohan:2067$ 

The load doesn't use sp directly.  There are lots of different ways that sp can be used in a function.  We need something more general to check for all sp uses and adjust them, or disable the optimization to remove the save/restore calls if too complicated.

There are some minor coding style issues with the patch.  Two comments don't end with a period and two spaces, and some of the indentation might not be quite right, but it is hard to tell looking at a patch in github.  This is stuff that can be fixed later though.

fakepaper56:
Hi all,
I rewrite the function update_sp_use(). The new patch consider use of sp has 3 ways:
1. load stack incomming argument.
2. load stack incomming argument and sign/zero-extend it.
3. regs_not_sp = sp + offset.

But I have no idea that "regs_not_sp = sp + reg" exists?

I does the following experiment:
$ cat sample.c
int foo(char *);
int bar(int a, int b, int c, int d, int e, int f, int g, int h, int u) {
  char *ptr = (char *)&u + a;
  return foo(ptr);
}
$ riscv32-unknown-elf-gcc -S -msave-restore -Os temp.c -da -dp

The 2 files shows the problem is related with reload pass.

temp.c.283r.ira:
(insn 14 13 15 2 (set (reg:SI 10 a0)
        (plus:SI (reg/f:SI 64 arg)
            (reg:SI 85))) "temp.c":4:8 3 {addsi3}
     (expr_list:REG_DEAD (reg:SI 85)
        (nil)))

temp.c.284r.reload:
(insn 25 13 14 2 (set (reg:SI 15 a5 [86])
        (plus:SI (reg/f:SI 2 sp)
            (const_int 16 [0x10]))) "temp.c":4:8 3 {addsi3}
     (nil))

(insn 14 25 15 2 (set (reg:SI 10 a0)
        (plus:SI (reg:SI 15 a5 [86])
            (reg:SI 10 a0 [85]))) "temp.c":4:8 3 {addsi3}
     (nil))

It seems I at least need to know how reload do...
I'm glad to survey reload implementation, but I have no idea how long I could finish the task.

kito-cheng:
Maybe we could stop optimizing save restore call for those function as temporary work-around solution?

```c
diff --git a/gcc/config/riscv/riscv-sr.c b/gcc/config/riscv/riscv-sr.c
index 9af50ef028e3..6d3dfe63a704 100644
--- a/gcc/config/riscv/riscv-sr.c
+++ b/gcc/config/riscv/riscv-sr.c
@@ -244,6 +244,9 @@ check_for_no_return_call (rtx_insn *prologue)
 void
 riscv_remove_unneeded_save_restore_calls (void)
 {
+  if (crtl->args.size == 0)
+    return;
+
   /* Will point to the first instruction of the function body, after the
      prologue end note.  */
   rtx_insn *body = NULL;

```

jim-wilson:
Disabling the optimization for this case is the easy short term solution.  And then try to put a better fix into gcc 11.

jim-wilson:
Before reload, we have a fake arg pointer register that points at incoming arguments on the stack.  During reload, we do register elimination which replaces the fake arg pointer register with sp + offset.  The offset depends on the frame size, so it can't be computed until we finalize register allocation, which happens in reload.  See ELIMINABLE_REGS and INITIAL_ELIMINATION_OFFSET in gcc/config/riscv/riscv.h and the code that uses these macros.  We also have a fake soft frame pointer.  The arg and soft-frame pointers will be eliminated to either sp or (hard) fp depending on whether the hard fp exists, and whether the offset to sp is constant or not, and maybe other factors.  If you look at the reload dump, you can see lines like
Can eliminate 64 to 2 (offset=16, prev_offset=0)
where 64 is the fake arg pointer register, and 2 is sp.

So before reload, we have an address arg + a0, and during reload that gets expanded to sp + 16 + a0 because the frame size is 16 bytes, which then has to be canonicalized because we don't have an add that takes 3 inputs, so reload fixes this by changing it to tmp = sp + 16; tmp = tmp + a0.  There isn't anything wrong with reload or the testcase here.  There isn't any other way to generate correct code.  We just need to handle all possible uses of sp correctly.

I would suggest using -fdump-rtl-all instead of -da.  -da is the old option name.

gcc/config/riscv/riscv-sr.c
```diff
@@ -171,6 +171,42 @@ riscv_sr_match_epilogue (void)
...
+  for (insn = NEXT_INSN (prologue); insn != epilogue; insn = NEXT_INSN (insn))
+    if (INSN_P (insn) && GET_CODE (PATTERN (insn)) == SET)
+      {
+	rtx pat_src = SET_SRC (PATTERN (insn));
+
+	/* case assigning sp value */
+	if (GET_CODE (pat_src) == PLUS && REG_P (XEXP (pat_src, 0))
+	    && REGNO (XEXP (pat_src, 0)) == STACK_POINTER_REGNUM)
+	  {
+	    INTVAL (XEXP (pat_src, 1)) -= TARGET_RVE ? 12 : 16;
```
jim-wilson, line 194: Doesn't verify that XEXP (pat_src, 1) is a CONST_INT before modifying it.

gcc/config/riscv/riscv-sr.c
```diff
@@ -171,6 +171,42 @@ riscv_sr_match_epilogue (void)
...
+
+	/* case loading stack incoming arguments  */
+	if (GET_CODE (pat_src) == SIGN_EXTEND
+	    || GET_CODE (pat_src) == ZERO_EXTEND)
+	  pat_src = XEXP (pat_src, 0);
+
+	if (MEM_P (pat_src) && GET_CODE (XEXP (pat_src, 0)) == PLUS
+	    && REG_P (XEXP (XEXP (pat_src, 0), 0))
+	    && REGNO (XEXP (XEXP (pat_src, 0), 0)) == STACK_POINTER_REGNUM)
+	  INTVAL (XEXP (XEXP (pat_src, 0), 1)) -= TARGET_RVE ? 12 : 16;
```
jim-wilson, line 206: Doesn't verify that XEXP (XEXP (pat_src, 0), 1) is a CONST_INT before modifying it.

jim-wilson:
I prefer Kito's suggested change.  It is simpler and safer.  The code size and performance loss here should be negligible.  Though Kito's patch probably should have a comment explaining that sp offsets will be wrong for eliminated arg pointer references which is why we can't optimize this case.  And maybe add the testcases too in case we want to try to optimize this case later.

fakepaper56:
I prefer Kito's change too. And I thank for jim's review and suggestion.

kito-cheng:
Fixed on upstream.

https://github.com/gcc-mirror/gcc/commit/4c0d1322033ce979532425d336530b217f6b5fd3

PR #199, opened 2020-10-01
Fix build failure with zstd version 1.2.0 or older.

jim-wilson:
Extends the configure check for zstd.h to also verify the zstd version,
since gcc requires features that only exist in 1.3.0 and newer.  Without
this patch we get a build error for lto-compress.c when using an old zstd
version.

	Backported from master:
	2020-09-29  Jim Wilson  <jimw@sifive.com>

	gcc/
	PR bootstrap/97183
	* configure.ac (gcc_cv_header_zstd_h): Check ZSTD_VERISON_NUMBER.
	* configure: Regenerated.

### DIFF
--- a/gcc/configure
+++ b/gcc/configure
@@ -10024,9 +10024,14 @@ $as_echo_n "checking for zstd.h... " >&6; }
 if ${gcc_cv_header_zstd_h+:} false; then :
   $as_echo_n "(cached) " >&6
 else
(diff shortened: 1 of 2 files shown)


jim-wilson:
This is for riscv-gnu-toolchain #718.

kito-cheng:
Thanks!

PR #200, opened 2020-10-20
Minimally tested sb*iw support.

jim-wilson:


aswaterman:
Thanks for quickly putting this together, @jim-wilson!

kito-cheng (approved):
Verified with following testcase:

```
int foo1(int x, int y){
    return (x + y)| 0x8000;
}
int foo2(int x, int y){
    return (x + y) ^ 0x8000;
}
int foo3(int x, int y){
    return (x + y) & (~0x8000);
}
```

gcc/config/riscv/bitmanip.md
```diff
@@ -153,6 +153,14 @@
   "sbsetw\t%0,%1,%2"
   [(set_attr "type" "bitmanip")])
 
+(define_insn "*sbsetiw"
+  [(set (match_operand:DI 0 "register_operand" "=r")
+	(ior:DI (sign_extend:DI (match_operand:SI 1 "register_operand" "=r"))
```
kito-cheng, line 158: Redundant `=` for input operand, which is reported by Evgeny before: https://github.com/riscv/riscv-bitmanip/issues/79

kito-cheng:
I also created another branch based on gcc 10.2 https://github.com/riscv/riscv-gcc/tree/riscv-gcc-10.2.0-rvb , and I will run test with that branch :)

jim-wilson:
I fixed the =r input operatnd problems in my patch and in the surrounding patterns.

jim-wilson:
Sorry, I didn't read that clearly enough.  I didn't realize you already fixed the =r problem on the new branch.  I can move to the new branch and submit a pr there.

PR #204, opened 2020-10-31
Fix rotate modes, must be QI same as shifts.

jim-wilson:
/* Expect li -2 and rol.  */
/* Or sbsetw and not.  */
long
li_rol_or_sbset_not (long i)
{
  return ~(1L << i);
}

Without rotate mode fix, we get
li_rol_or_sbset_not:
        li      a5,1
        sll     a0,a5,a0
        not     a0,a0
        ret
with rotate mode fix we get
li_rol_or_sbset_not:
        li      a5,-2
        rol     a0,a5,a0
        ret

Also sbset/not should be option if no rol, but there is a pttern missing.  That is my next patch.


kito-cheng (approved):
Verified with the testcase :)

PR #205, opened 2020-10-31
Add sbset with constant 1 input.

jim-wilson:
Testcase is same as the rotate mode fix, except now for
long
li_rol_or_sbset_not (long i)
{
  return ~(1L << i);
}
we get if Zbs enabled
li_rol_or_sbset_not:
        sbset   a0,x0,a0
        not     a0,a0
        ret


kito-cheng (approved):
Verified with testcase and GCC testsuite.

PR #206, opened 2020-10-31
Use li and rori to load constants.

jim-wilson:
I doubt that this will help many applications, the constants are a bit too obscure.  But we may as well try to get these cases right just in case.

This only handles the li negative followed by rori case.  There are a few more cases we could handle.  li positive followed by rori is useful for the case where the ones wrap around.  For instance 0xf00000000000000f could be loaded with li 255; rori 4.  We can also load some constants using lui followed by rori.  There may also be other cases I haven't thought of yet.

There are other ways to implement this.  I haven't given this much thought yet.  For instance, we could call ctz, then rotate by that amount, and then check for a SMALL_INT constant.  That might be easier to understand than the current code.  That doesn't work for the wraparound case, but we could rotate by 32, call ctz, rotate again, and then check for a SMALL_INT.

### DIFF
--- a/gcc/config/riscv/riscv.c
+++ b/gcc/config/riscv/riscv.c
@@ -455,6 +455,47 @@ riscv_build_integer_1 (struct riscv_integer_op codes[RISCV_MAX_INTEGER_OPS],
 	}
     }
 
+  if (cost > 2 && TARGET_64BIT && TARGET_BITMANIP)
+    {
+      int leading_ones = clz_hwi (~value);
+      int trailing_ones = ctz_hwi (~value);
+
+      /* If all bits are one except a few that are zero, and the zero bits
+	 are within a range of 11 bits, and at least one of the upper 32-bits
+	 is a zero, then we can generate a constant by loading a small
+	 negative constant and rotating.  */
+      if (leading_ones < 32
+	  && ((64 - leading_ones - trailing_ones) < 12))
+	{
+	  codes[0].code = UNKNOWN;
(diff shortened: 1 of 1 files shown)


jim-wilson:
Example testcases

Example testcases
/* Expect li -126 and rori.  */
long
li_rori (void)
{
  return 0xffff77ffffffffffL;
}

long
li_rori_2 (void)
{
  return 0x77ffffffffffffffL;
}

long
li_rori_3 (void)
{
  return 0xfffffffeefffffffL;
}

long
li_rori_4 (void)
{
  return 0x5ffffffffffffff5L;
}

long
li_rori_5 (void)
{
  return 0xaffffffffffffffaL;
}

Just trying to handle cases where we wraparound, cross the bit 32-31 boundary, have zero bits in the sign bit and/or the low bit.

kito-cheng:
I guess we can optimize constant value with slo and sro too, but it out of scope for this patch, and I know we are focusing on zbb, zba and zbs only now :P

kito-cheng (approved):
LGTM, verified with provided testcase and GCC testsuite.

jim-wilson:
FYI I put the new code at the end of riscv_build_integer_1 thinking that we should only use B instructions if we can't use base instructions, but I think that was a mistake, because putting it after the recursive calls means we can't use the rori trick in 3-4 intruction sequences.  So we should also try moving this before the recursive calls.  In addition to handling other cases like lui and positive constants rotated.

PR #208, opened 2020-11-03
Fix slliuw support, disable non-B splitter.

jim-wilson:
Fix issue documented in riscv-bitmanip issue 65.  The testcase is:

long
slliuw (long i)
{
  return (long)(unsigned int)i << 10;
}

which should generate a slliuw instruction when the B extension is enabled.


kito-cheng (approved):
LGTM, verified with rv64gcb

PR #209, opened 2020-11-03
Store large value in register, when value is out of range in a ADD instruction

monkchiang:
If offset is out of range in a  ADDI instruction. So Sp will not insert a instruction to adjust Sp value.

The testcase is:

int main(void)
{
  int *p;
  int i[2000];
  p = __builtin_alloca(2);

  return 0;
}

### DIFF
--- a/gcc/config/riscv/riscv.c
+++ b/gcc/config/riscv/riscv.c
@@ -4511,8 +4511,7 @@ riscv_expand_epilogue (int style)
       poly_int64 adjust = -frame->hard_frame_pointer_offset;
       rtx adjust_rtx = NULL_RTX;
 
-      if (!adjust.is_constant ()
-	  || (adjust.is_constant () && !SMALL_OPERAND (adjust.is_constant())))
+      if (!adjust.is_constant ())
 	{
 	  rtx tmp1 = RISCV_PROLOGUE_TEMP (Pmode);
 	  rtx tmp2 = RISCV_PROLOGUE_TEMP2 (Pmode);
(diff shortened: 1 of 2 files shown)


kito-cheng:
Thanks for fixing this! could you add a run-able testcase for that?

monkchiang:
Here is tesecase, for checking SP value whether to restore in epilogue.

jim-wilson:
I don't understand what this is fixing.  The if statement does
      if (!adjust.is_constant ()
	  || (adjust.is_constant () && !SMALL_OPERAND (adjust.is_constant())))
so how can we get to the else with a value that isn't a SMALL_OPERAND?

How exactly does the testcase fail?  Are you getting an ICE or what?

I do see a problem with the testcase.  We are subtracting 16 from the sp in the body of the function for the alloca, and then never adding it back.  This can only work if you have a frame pointer, and restore sp from the frame pointer at the end of the function.  We do have a frame pointer, but aren't using it in the epilogue.  Upstream mainline is using s0 to restore the stack in the epilogue.  The code is badly optimized, but looks correct.  We compute sp=s0-32 by a round about fashion, load the old value of s0, and then add 32 to sp.
	li	t1,-8192
	addi	t1,t1,160
	add	sp,s0,t1
	li	t1,8192
	addi	t1,t1,-192
	add	sp,sp,t1
	ld	s0,24(sp)
	addi	sp,sp,32
whereas on this rvv branch we have
	li	t1,8192
	addi	t1,t1,-192
	add	sp,sp,t1
	ld	s0,24(sp)
	addi	sp,sp,32
which is wrong because it isn't using s0 to compute the sp before the s0 restore.

jim-wilson:
The bad optimization is probably due to the -O0 which I forgot I was using.

jim-wilson:
Looks like the problem is the if statement.  The expression
    !SMALL_OPERAND (adjust.is_constant())
needs to be instead
   !SMALL_OPERAND (adjust.to_constant ())

note the change "is" to "to"

I probably would have called that constant_p instead of is_constant, which makes the mistake easier to see.

jim-wilson:
That fix leads to a different problem that the code calls riscv_gen_load_poly_int which assumes that the adjust value is not a constant.  So the if statement needs to be split into two, one with a non-constant adjust that calls riscv_gen_load_poly_int, and then a test for !SMALL_INT which does what Monk wrote, and then the else as currently for the SMALL_INT case.

monkchiang:
Thanks Jim point out another issue, I will modify the if statement. 
The second condition, I will remove it. (`(adjust.is_constant () && !SMALL_OPERAND (adjust.is_constant())`)
I will run some vector testcase to verify toolchain. And updating patch for this issue.

monkchiang:
I updated patch, after I verify it on vector and GCC testsuite.

jim-wilson (approved):
You could use an else if and eliminate a pair of braces, but nothing really wrong with writing it this way.

PR #216, opened 2020-11-23
sbset patterns with masked shift count.

jim-wilson:
Requires the previous patch to work.

We already have patterns for shifts with masked shift counts.  This does the same thing for sbset.  This fixes some minor code size regressions with sbset, where the andi gets optimized away without B but remained with B.

sbset is deliberately not part of the testcase file name or function names as that screws up the scan-assembler-times count.

Similar patches for clr, inv, and ext will follow.


jim-wilson:
FYI, this was working for me yesterday, but now I'm getting some odd linker errors.
/home/jimw/tmp-bitmanip/X-unpatched/build-install/lib/gcc/riscv32-unknown-elf/10.2.0/../../../../riscv32-unknown-elf/bin/ld: error: /home/jimw/tmp-bitmanip/X-unpatched/build-install/lib/gcc/riscv32-unknown-elf/10.2.0/crtbegin.o: Mis-matched ISA version for 'm' extension. 2.0 vs 2.0
The real problem here is that one file has b v2.0 and one file has b v0.0.  Binutils is just including the wrong extension.  The problem is in riscv_merge_std_ext where it uses find_in and find_out to do the version number comparion, but then passes in and out to riscv_version_mismatch.  in and out are pointing at the first extension, which is 'm'.  I think it should pass find_in and find_out to riscv_version_mismatch.  As for why the two files have different b version numbers, I haven't looked at that yet.

kito-cheng:
> The real problem here is that one file has b v2.0 and one file has b v0.0. Binutils is just including the wrong extension. The problem is in riscv_merge_std_ext where it uses find_in and find_out to do the version number comparion, but then passes in and out to riscv_version_mismatch. in and out are pointing at the first extension, which is 'm'. I think it should pass find_in and find_out to riscv_version_mismatch. As for why the two files have different b version numbers, I haven't looked at that yet.

Oh, I am resolving this issue recently, the root cause is GCC will always emit 2.0 for unknown extensions, but binutils will emit 0.0 for unknown extensions, so .c is b 2.0 and .s is b 0.0...

So the correct way should be fix both side, binutils and gcc should emit 0.92 for B, that's the reason why I summit version controlling patch [1].

[1] https://gcc.gnu.org/pipermail/gcc-patches/2020-November/558908.html

kito-cheng:
The wrong error message should fixed in upstream by Nelson, https://github.com/bminor/binutils-gdb/commit/32f0ce4db98eb1e8e1cc326d3a22ef0c436ae6fe

kito-cheng:
Verified with GCC testsuite.

PR #217, opened 2020-11-23
[RVB] Handle sub-extension and version info for B

kito-cheng:
This patch-set enable the separated switch for each sub-extensions, and emit the right version info for all B-extension family, also back-port 3 necessary patch from upstream.

This PR and binuitls's PR https://github.com/riscv/riscv-binutils-gdb/pull/239 should made everything work correctly.

### DIFF
--- a/gcc/common/config/riscv/riscv-common.c
+++ b/gcc/common/config/riscv/riscv-common.c
@@ -153,6 +153,129 @@ riscv_subset_list::~riscv_subset_list ()
     }
 }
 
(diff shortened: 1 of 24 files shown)


jim-wilson (approved):

jim-wilson:
I think the reason why this stuff worked for me over the weekend is that I didn't actually do any linking.  I was just disassembling object files to look for optimization opportunities, and running compile only testsuite testcases.  I can link now with this patch and the binutils patch I already merged in.

PR #218, opened 2020-11-24
Fixing vmv_s_x/vmv_x_s for rv32gcv

kito-cheng:
Fix https://github.com/riscv/riscv-gnu-toolchain/issues/779, not sure why we put the TARGET_64BIT in the pattern condition, 
I guess one possible is we don't know how to handle 64-bits vector-scalar operations yet, but block all ELEN would be too strong.

So let remove the TARGET_64BIT check at this moment.


MarkHillHuawei:
Thanks Kito-Cheng, I have applied the patch, rebuilt the rv32 toolchain and can confirm the correct code is now generated.

jim-wilson (approved):

PR #222, opened 2020-12-12
Update rvb support.

jim-wilson:
Remove some dropped insns.  Fix minor typos with subsets.

### DIFF
--- a/gcc/common/config/riscv/riscv-common.c
+++ b/gcc/common/config/riscv/riscv-common.c
@@ -985,7 +985,7 @@ static const riscv_ext_flag_table_t riscv_ext_flag_table[] =
   {"zbb", &gcc_options::x_riscv_bitmanip_subext, MASK_ZBB},
   {"zbs", &gcc_options::x_riscv_bitmanip_subext, MASK_ZBS},
   {"zbp", &gcc_options::x_riscv_bitmanip_subext, MASK_ZBP},
-  {"zbr", &gcc_options::x_riscv_bitmanip_subext, MASK_ZBP},
+  {"zbr", &gcc_options::x_riscv_bitmanip_subext, MASK_ZBR},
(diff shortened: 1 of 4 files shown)


jim-wilson:
The slliu.w testcase deserves mention.  The middle-end produces
  _1 = i_2(D) << 10;
  _3 = _1 & 4398046510080;
The constant is 0x3fffffffc00.  Without the zbs extension, it takes 3 instructions to load the constant, and then we need to combine 5 instructions to create the slliu.w which gcc doesn't do.  With the zbs extension, it takes 2 insns and a 4->1 combine, so it works.

Note that if we still had "addiwu a0,x0,-1" then we can load the constant with two instructions with addiwu and a shift.  Not that this is a good reason to re-add addiwu.  Just an observation that without zbs it might be useful if we had an instruction that can load 0x00000000ffffffff into a register, with just one instruction.

Alternatively, we may need to look for a way to fix the middle-end optimizer not to produce such inconvenient IL.  I think someone else (Philipp Tomsich?) may have run into the same issue recently.

kito-cheng:
I gonna merge this first and then I think we can keep improve code gen, and does this case same as the Philipp Tomsich[1] posted in mailing list?

[1] https://www.mail-archive.com/gcc-patches@gcc.gnu.org/msg248732.html

jim-wilson:
Yes, that patch from Philipp looks like the same problem.  Seems like the disussion ended with a question about the RISC-V constant cost model that wasn't answered.  I also pointed out that the problem could be fixed with a 4->3 splitter which combine currently doesn't support.  But I did see a patch for that from IBM
  https://gcc.gnu.org/pipermail/gcc-patches/2020-November/560573.html
This came after entering stage3, so I don't think anything will happen until spring.  It is possible that this patch plus a new combiner pattern could fix the problem.

PR #247, opened 2021-03-19
Improve cmix support, handle switched operands.

jim-wilson:
Given this testcase
long cmix (long i, long j, long k) { return (i & j) | (k & ~j); }
long cmix2 (long i, long j, long k) { return i ^ (j & (i ^ k)); }
long cmix3 (long i, long j, long k) { return i ^ (j & (k ^ i)); }
and compiling with zbt enabled we only generate cmix for 2 of the 3 functions.  This patch gives us cmix for all 3 testcases.

Minimally tested.

### DIFF
--- a/gcc/config/riscv/bitmanip.md
+++ b/gcc/config/riscv/bitmanip.md
@@ -385,6 +385,18 @@
   "cmix\t%0,%2,%1,%3"
   [(set_attr "type" "bitmanip")])
 
+;; ??? Can we do this by using the % communtative constraint?
+
+(define_insn "*cmix2"
+  [(set (match_operand:X 0 "register_operand" "=r")
+	(xor:X (and:X (xor:X (match_operand:X 1 "register_operand" "r")
+			     (match_operand:X 3 "register_operand" "r"))
+		      (match_operand:X 2 "register_operand" "r"))
+	       (match_dup 1)))]
+  "TARGET_ZBT"
+  "cmix\t%0,%2,%3,%1"
+  [(set_attr "type" "bitmanip")])
+
 ;;; ??? cmov
 
 ;;; ??? fs[lr]


kito-cheng (approved):
Verified with the testcase :)

gcc/config/riscv/bitmanip.md
```diff
@@ -385,6 +385,18 @@
...
   [(set_attr "type" "bitmanip")])
 
+;; ??? Can we do this by using the % communtative constraint?
+
+(define_insn "*cmix2"
+  [(set (match_operand:X 0 "register_operand" "=r")
+	(xor:X (and:X (xor:X (match_operand:X 1 "register_operand" "r")
+			     (match_operand:X 3 "register_operand" "r"))
+		      (match_operand:X 2 "register_operand" "r"))
+	       (match_dup 1)))]
```
kito-cheng, line 395: Maybe use ` (match_operand:X 4 "register_operand" "1,3")`  here and then we could merged with `*cmix` pattern?

jim-wilson:
Use of match_operand 1,3 doesn't work, as reload will force the instruction to fit the constraints, and there is no way to reload two different input operands into the same register.

Given
long cmix4 (long i, long j, long k, long l) { return l ^ (j & (k ^ i)); }
I get
cmix4:
	cmix	a0,a1,a0,a2
	ret
which is wrong as we lost the a3 reference.

I get in the 283. ira file I have
(insn 16 22 17 2 (set (reg/i:DI 10 a0)
        (xor:DI (and:DI (xor:DI (reg:DI 84)
                    (reg:DI 82))
                (reg:DI 83))
            (reg:DI 85))) "tmp.c":1:73 332 {*cmix0}
     (expr_list:REG_DEAD (reg:DI 85)
        (expr_list:REG_DEAD (reg:DI 84)
            (expr_list:REG_DEAD (reg:DI 83)
                (expr_list:REG_DEAD (reg:DI 82)
                    (nil))))))
whcih is OK, but then in 284 reload I have
(insn 16 22 17 2 (set (reg/i:DI 10 a0)
        (xor:DI (and:DI (xor:DI (reg:DI 12 a2 [84])
                    (reg:DI 10 a0 [82]))
                (reg:DI 11 a1 [83]))
            (reg:DI 12 a2 [84]))) "tmp.c":1:73 332 {*cmix0}
     (nil))
which is wrong because reload put two different input operands in the same register.

PR #250, opened 2021-03-30
[crypto] add instructions in K-ext spec v0.9.0

WuSiYu:
add instructions (and their builtins) in K-ext spec v0.9.0, except those in Zkb and Zkg which is waiting for ext-b's work to prevent conflict.

### DIFF
--- a/gcc/common/config/riscv/riscv-common.c
+++ b/gcc/common/config/riscv/riscv-common.c
@@ -237,7 +237,7 @@ riscv_subset_list::lookup (const char *subset, int major_version,
 static const char *
 riscv_supported_std_ext (void)
 {
-  return "mafdqlcbjtpvn";
+  return "mafdqlcbjktpvn";
 }
 
 /* Parsing subset version.
@@ -589,6 +589,12 @@ riscv_subset_list::parse (const char *arch, location_t loc)
   if (p == NULL)
     goto fail;
 
+  /* Parsing crypto extension.  */
+  p = subset_list->parse_multiletter_ext (p, "k", "crypto extension");
+
+  if (p == NULL)
+    goto fail;
+
   /* Parsing sub-extensions.  */
   p = subset_list->parse_multiletter_ext (p, "z", "sub-extension");
 
@@ -645,6 +651,19 @@ static const riscv_ext_flag_table_t riscv_ext_flag_table[] =
   {"f", &gcc_options::x_target_flags, MASK_HARD_FLOAT},
   {"d", &gcc_options::x_target_flags, MASK_DOUBLE_FLOAT},
   {"c", &gcc_options::x_target_flags, MASK_RVC},
+
+  {"k", &gcc_options::x_riscv_crypto_subext, MASK_ZKNE | MASK_ZKND | MASK_ZKNH | MASK_ZKG | MASK_ZKB | MASK_ZKR},
+  {"zkg", &gcc_options::x_riscv_crypto_subext, MASK_ZKG},
+  {"zkb", &gcc_options::x_riscv_crypto_subext, MASK_ZKB},
+  {"zkr", &gcc_options::x_riscv_crypto_subext, MASK_ZKR},
+  {"zkn", &gcc_options::x_riscv_crypto_subext, MASK_ZKNE | MASK_ZKND | MASK_ZKNH | MASK_ZKG | MASK_ZKB},
+  {"zkne", &gcc_options::x_riscv_crypto_subext, MASK_ZKNE},
+  {"zknd", &gcc_options::x_riscv_crypto_subext, MASK_ZKND},
+  {"zknh", &gcc_options::x_riscv_crypto_subext, MASK_ZKNH},
+  {"zks", &gcc_options::x_riscv_crypto_subext, MASK_ZKSED | MASK_ZKSH | MASK_ZKG | MASK_ZKB},
+  {"zksed", &gcc_options::x_riscv_crypto_subext, MASK_ZKSED},
+  {"zksh", &gcc_options::x_riscv_crypto_subext, MASK_ZKSH},
+
   {NULL, NULL, 0}
 };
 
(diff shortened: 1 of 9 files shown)


gcc/common/config/riscv/riscv-common.c
```diff
@@ -645,6 +651,19 @@ static const riscv_ext_flag_table_t riscv_ext_flag_table[] =
   {"f", &gcc_options::x_target_flags, MASK_HARD_FLOAT},
   {"d", &gcc_options::x_target_flags, MASK_DOUBLE_FLOAT},
   {"c", &gcc_options::x_target_flags, MASK_RVC},
+
+  {"k", &gcc_options::x_riscv_crypto_subext, MASK_ZKNE | MASK_ZKND | MASK_ZKNH | MASK_ZKG | MASK_ZKB | MASK_ZKR},
```
kito-cheng, line 655: I would suggest you using `riscv_implied_info` to record the implied extension info like `zb*`, you could reference her e: https://github.com/riscv/riscv-gcc/blob/riscv-gcc-10.2.0-rvb/gcc/common/config/riscv/riscv-common.c#L63

gcc/common/config/riscv/riscv-common.c
```diff
@@ -645,6 +651,19 @@ static const riscv_ext_flag_table_t riscv_ext_flag_table[] =
...
+  {"zkg", &gcc_options::x_riscv_crypto_subext, MASK_ZKG},
+  {"zkb", &gcc_options::x_riscv_crypto_subext, MASK_ZKB},
+  {"zkr", &gcc_options::x_riscv_crypto_subext, MASK_ZKR},
+  {"zkn", &gcc_options::x_riscv_crypto_subext, MASK_ZKNE | MASK_ZKND | MASK_ZKNH | MASK_ZKG | MASK_ZKB},
```
kito-cheng, line 659: Ditto.

gcc/common/config/riscv/riscv-common.c
```diff
@@ -645,6 +651,19 @@ static const riscv_ext_flag_table_t riscv_ext_flag_table[] =
...
+  {"zkne", &gcc_options::x_riscv_crypto_subext, MASK_ZKNE},
+  {"zknd", &gcc_options::x_riscv_crypto_subext, MASK_ZKND},
+  {"zknh", &gcc_options::x_riscv_crypto_subext, MASK_ZKNH},
+  {"zks", &gcc_options::x_riscv_crypto_subext, MASK_ZKSED | MASK_ZKSH | MASK_ZKG | MASK_ZKB},
```
kito-cheng, line 663: Ditto.

kito-cheng (commented):
I would suggest you add at least one test case for each sub-extension of intrinsic functions, to make sure the code gen is expected.

gcc/config/riscv/crypto.md
```diff
@@ -0,0 +1,302 @@
...
+        (match_operand:SI 2 "immediate_operand" "")]
+        UNSPEC_CRYPTO))]
+  "TARGET_ZKSED"
+  "sm4ks\t%0,%1,%2")
+
+
+;; Zkr - Entropy Source
+
+(define_insn "riscv_pollentropy_<mode>"
+  [(unspec_volatile [(match_operand:X 0 "register_operand" "=r")] UNSPEC_CRYPTO)]
```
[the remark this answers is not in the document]
WuSiYu: I tried to changed this to `(set (match_operand:X 0 "register_operand" "=r") (unspec:X UNSPEC_XXX))`, but it seem that it can't pass the compile, the "unspec" wants a vector after it.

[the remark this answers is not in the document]
kito-cheng: Use a dummy `const_int 0` for that, like  `(set (match_operand:X 0 "register_operand" "=r") (unspec:X [(const_int 0)] UNSPEC_XXX))`, this trick are used in several different back-end :)

You can grep by following command to see:
```
grep "\[(const_int 0)\]" * -R |grep unspec
```

gcc/config/riscv/crypto.md
```diff
@@ -0,0 +1,302 @@
+;; Machine description for K extension.
+;; Copyright (C) 2011-2021 Free Software Foundation, Inc.
```
jim-wilson, line 2: This is a new file.  It only needs to mention the current year.

gcc/config/riscv/crypto.md
```diff
@@ -0,0 +1,302 @@
...
+;; You should have received a copy of the GNU General Public License
+;; along with GCC; see the file COPYING3.  If not see
+;; <http://www.gnu.org/licenses/>.
+
+
+;; Zkne&Zknd - AES (RV32)
+
+(define_insn "riscv_aes32dsi"
+  [(set (match_operand:SI 0 "register_operand" "=r")
+    (unspec:SI
```
jim-wilson, line 25: If you are an emacs user, you can do M-x lisp-mode and then tab will indent the RTL properly.  The open parenthesis here should align with the (match_operand on the line above.

jim-wilson (commented):
Just some minor comments to add to a lot of good comments from Kito.

WuSiYu:
Thanks for your comments and guidance! I pushed a new commit to fix these. (And to make the `riscv_implied_info` works, I also cherry-pick a patch from `riscv-gcc-10.2.0-bump` branch.)
I think I fixed most of them, but there are still some questions that may need your guidance.

@kito-cheng For the UNSPEC_* part, I split them into the things like `UNSPEC_AES_DS`, `UNSPEC_AES_DS` ... ,I am not sure if this is too sketchy or too detailed? please tell me if I did it wrong.

@jim-wilson Unfortunately I am not familiar with emacs, but I found a tool that use emacs to format lisp code, I use it on `crypto.md` and hope it worked fine for the .md.

kito-cheng:
This patch is LGTM now.
One minor suggestion is you could using `define_int_iterator` + `define_int_attr` to further simplify RTL pattern, but it's not must to do.

PR #251, opened 2021-04-02
Update 2 C++ coroutine testcases from upstream.

jim-wilson:
Fixes use-after-free bugs exposed by new glibc version.


kito-cheng:
Thanks!

PR #257, opened 2021-04-18
RISC-V: Fix testcase of vector to be compatible with intrinsic API (#…

rjiejie:


rjiejie:
@kito-cheng any comment for this patch ?

kito-cheng:
Thanks, @rjiejie , I've verified the change, it's really great job :)

PR #260, opened 2021-05-07
RISC-V: Fix unsupported insn 'fabs.h' based on ZFH v0.1

rjiejie:
The spec of ZFH does not support instruction 'fabs.h', we need to delete unsupported instruction
and keep compatible with binutils :)

### DIFF
--- a/gcc/config/riscv/riscv.md
+++ b/gcc/config/riscv/riscv.md
@@ -390,6 +390,9 @@
 			    (DF "TARGET_DOUBLE_FLOAT")
 			    (HF "TARGET_FP16")])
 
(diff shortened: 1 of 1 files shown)


aswaterman:
This sounds like a binutils bug: fabs.h is just an alias for fsgnx.h.

Regardless, this doesn’t seem to be the right fix. If this indeed a problem, we should work around it, rather than just deleting the patterns from GCC.

kito-cheng:
I am prefer fix binutils rather than drop this pattern is GCC too.

rjiejie:
@aswaterman @kito-cheng Okay, it looks better choice :)

PR #262, opened 2021-05-16
[bitmanip][WiP] [RFC] Add automatic generation of pack*

rdolbeau:
This adds automatic generation of pack* instructions (pack, packu, packh) beyond zero-extension.
Unfortunately it's very ad-hoc at the moment.

I'm using the peephole2 pass, but it seems very restrictive in what it matches and perhaps not the best solution.
Any suggestion of where/how to implement this properly welcome.

This has not been tested very much and might be buggy, this is for discussion on how to best leverage B in GCC.
[edit] also currently it's really only for RV32.

The attached pack-pattern.txt is a C file with some C code pattern that should be recognized to generate pack[,u,h]. Some (not all) are identified by this patch. Some of those patterns are identified in real-life codes and the appropriate instruction[s] is/are generated (with RV32IMAFDCK_ZBR_ZBT as the march), such as in Cairo 1.17.4 (`cairo-1.17.4/src/cairo-image-info.c`) or Poppler 21.05.0 (`poppler-21.05.0/fofi/FoFiTrueType.cc`) - both are mostly packh/packh/pack to assemble 4 bytes in a word (matching a *very* ad-hoc pattern...).

@clairexen any suggestion on what packages should have pack*-suitable code patterns? [edit] for rv32; the Skia examples use uint32_t so only need pack* in rv64.

[edit] updated version of the pattern file (4x)
[pack-pattern.txt](https://github.com/riscv/riscv-gcc/files/6509967/pack-pattern.txt)

### DIFF
--- a/gcc/config/riscv/bitmanip.md
+++ b/gcc/config/riscv/bitmanip.md
@@ -448,6 +448,26 @@
 
 ;;; ??? cmov
 
+(define_insn "*mov<X:mode>cc_ne_bitmanip"
+  [(set (match_operand:X 0 "register_operand" "=r")
+	(if_then_else:X
+	 (ne (match_operand:X 1 "register_operand" "r") (const_int 0))
+	 (match_operand:X 2 "register_operand" "r")
+	 (match_operand:X 3 "register_operand" "r")))]
+  "TARGET_ZBT"
+  "cmov\t%0,%1,%2,%3"
+  [(set_attr "type" "bitmanip")])
+
+(define_insn "*mov<X:mode>cc_eq_bitmanip"
+  [(set (match_operand:X 0 "register_operand" "=r")
+	(if_then_else:X
+	 (eq (match_operand:X 1 "register_operand" "r") (const_int 0))
+	 (match_operand:X 2 "register_operand" "r")
+	 (match_operand:X 3 "register_operand" "r")))]
+  "TARGET_ZBT"
+  "cmov\t%0,%1,%3,%2"
+  [(set_attr "type" "bitmanip")])
+
 ;;; ??? fs[lr]
 
 (define_insn "*shNadd"
--- a/gcc/config/riscv/bitmanip.md
+++ b/gcc/config/riscv/bitmanip.md
@@ -83,6 +83,55 @@
  
 ;;; ??? pack
 
+(define_insn "riscv_bitmanip_packsi"
+ [(set (match_operand:SI 0 "register_operand" "=r")
+       (ior:SI (and:SI (match_operand:SI 1 "register_operand" "r") (const_int 65280))
+               (ashift:SI (match_operand:SI 2 "register_operand" "r") (const_int 16))))]
+  "TARGET_ZBP"
+  { return TARGET_64BIT ? "packw\t%0,%1,%2" : "pack\t%0,%1,%2"; }
+  [(set_attr "type" "bitmanip")
+   (set_attr "length" "4")])
+
+(define_insn "riscv_bitmanip_packdi"
+ [(set (match_operand:DI 0 "register_operand" "=r")
+       (ior:DI (and:SI (match_operand:DI 1 "register_operand" "r") (const_int 65280))
+               (ashift:DI (match_operand:DI 2 "register_operand" "r") (const_int 32))))]
+  "TARGET_ZBP && TARGET_64BIT"
+  "pack\t%0,%1,%2"
+  [(set_attr "type" "bitmanip")
+   (set_attr "length" "4")])
+
+(define_insn "riscv_bitmanip_packusi"
+ [(set (match_operand:SI 0 "register_operand" "=r")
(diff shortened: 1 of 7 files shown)


gcc/config/riscv/bitmanip.md
```diff
@@ -83,6 +83,74 @@
...
+(define_insn "*extractB0toB1"
+[(set (match_operand:SI 0 "register_operand" "=r")
+      (and:SI (ashift:SI (match_operand:SI 1 "register_operand" "r")
+	                     (const_int 8))
+              (const_int 65280)))]
+"!TARGET_64BIT"
+"andi\\t%0,%1,0xff\;slli\\t%0,%0,0x8"
+[(set_attr "length" "8")]
+)
+(define_insn "*extractB0toB2"
```
kito-cheng, line 121: I would prefer `define_split` or `define_insn_and_split` for `extractB0toB1` and `extractB0toB2`, although I think this should fixed in front-end or middle-end, but this fix is much simpler than modify front-end/middle-end,

gcc/config/riscv/peephole_pack.md
```diff
@@ -0,0 +1,587 @@
...
+  [(set (match_dup 0) (plus:SI (match_dup 2) (match_dup 1)))]
+{
+  emit_insn(gen_zero_extendqisi2(operands[3], operands[1]));
+  emit_insn(gen_zero_extendqisi2(operands[4], operands[2]));
+  emit_insn(gen_riscv_bitmanip_packhsi(operands[0], operands[4], operands[3])); 
+  DONE;
+})
+
+;; Does it look ad-hoc? I think it looks ad-hoc...
+(define_peephole2
```
kito-cheng, line 293: I believe those ad-hoc pattern could be removed once `pack*` pattern has describe right :)
Or...at least could be simplified.

kito-cheng (commented):
Do you or SiPearl has signed FSF copyright assignment? If not, could you signed that to prevent any copyright issue in future :)

rdolbeau:
@kito-cheng I originally tried without unspec, but couldn't get it to reliably match/be matched at the time, so fell back on a lot of unspec.

I've never worked extensively with GCC's MD, and the little I did was 15 years ago :-) I'm sure there's cleaner/better way to do this, which is the reason for this PR; a starting point to get pack* supported properly in GCC.

A lot of the 'ad-hoc-ness' comes from the need to match a lot of instructions to identify the pattern, to make sure only 8 bits are valid/used. The one case that currently doesn't work is uint8_t as argument to a function; as far as I can tell there's not way to tell in the back-end that the 3 MSbytes of the inputs are 0 - that information is already gone. Load bytes work, and seems to be what appear in actual libraries.

And no I don't have a copyright assignment with the FSF at this time.

rdolbeau:
@kito-cheng Something like 5fd0898 ? (only for UNSPEC_PACK ATM). But I still need the pack-generating peephole, otherwise the instruction is not generated...

gcc/config/riscv/bitmanip.md
```diff
@@ -83,6 +83,74 @@
...
+  "TARGET_ZBP"
+  "pack\t%0,%1,%2"
+  [(set_attr "type" "bitmanip")
+   (set_attr "length" "4")])
+(define_insn "riscv_bitmanip_packu<mode>"
+  [(set (match_operand:X 0 "register_operand" "=r")
+	(unspec
+	    [(match_operand:X 1 "register_operand" "r")
+		 (match_operand:X 2 "register_operand" "r")]
+	    UNSPEC_PACKU))]
```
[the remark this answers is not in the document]
rdolbeau: done

gcc/config/riscv/bitmanip.md
rdolbeau: done

gcc/config/riscv/bitmanip.md
```diff
@@ -83,6 +83,74 @@
  
 ;;; ??? pack
 
+(define_insn "riscv_bitmanip_pack<mode>"
+  [(set (match_operand:X 0 "register_operand" "=r")
+	(unspec
+	    [(match_operand:X 1 "register_operand" "r")
+		 (match_operand:X 2 "register_operand" "r")]
+	    UNSPEC_PACK))]
```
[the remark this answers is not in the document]
rdolbeau: done

gcc/config/riscv/peephole_pack.md
rdolbeau: Actually they still seem necessary...

rdolbeau:
@kito-cheng While the 'pack' stuff is too complicated for its own good, maybe c758e92 (which only implement support for `sext.[bh]`) is worth cherry-picking - and simple enough that the lack of copyright assignment is not a problem? (raised the subject internally, shouldn't be a problem, but might take a while...).

rdolbeau:
This has diverged quite a bit from the original code (from peepholes to a specific pass earlier in the process), I'll close it and reopen a new one.

PR #264, opened 2021-05-17
[crypto] Add testcase and some fix

WuSiYu:
Add testcase for RISC-V K-ext, and some fix for last K-ext.
Now all tests (except the old memcpy bug) in make report-gcc-newlib are passed.


kito-cheng:
This patch-set seems mixed-in some other patches, could you keep separated commit for those changes?

WuSiYu:
Sorry for late, the commits now has been split into fixs in K-ext, testcase for K-ext, and other patches.

PR #265, opened 2021-05-26
[bitmanip] Generate sext.[bh] for sign-extension under ZBB

rdolbeau:
This simply adds a new define_insn for sext.[bh] when ZBB is available (and deactivate the existing define_insn_and_split for sign_extend).

### DIFF
--- a/gcc/config/riscv/riscv.md
+++ b/gcc/config/riscv/riscv.md
@@ -1134,11 +1134,16 @@
   [(set_attr "move_type" "move,load")
    (set_attr "mode" "DI")])
 
-(define_insn_and_split "extend<SHORT:mode><SUPERQI:mode>2"
+(define_expand "extend<SHORT:mode><SUPERQI:mode>2"
+  [(set (match_operand:SUPERQI 0 "register_operand" "")
+    (sign_extend:SUPERQI
+	    (match_operand:SHORT 1 "nonimmediate_operand" "")))])
+
+(define_insn_and_split "*extend<SHORT:mode><SUPERQI:mode>2"
   [(set (match_operand:SUPERQI   0 "register_operand"     "=r,r")
 	(sign_extend:SUPERQI
 	    (match_operand:SHORT 1 "nonimmediate_operand" " r,m")))]
-  ""
+  "!TARGET_ZBB"
   "@
    #
    l<SHORT:size>\t%0,%1"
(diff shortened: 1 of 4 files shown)


kito-cheng:
@rdolbeau this patch is LGTM, could you sign FSF copyright assignment? otherwise it could be an issue when we try to upstream that.

gcc/config/riscv/riscv.md
```diff
@@ -1138,7 +1138,7 @@
   [(set (match_operand:SUPERQI   0 "register_operand"     "=r,r")
 	(sign_extend:SUPERQI
 	    (match_operand:SHORT 1 "nonimmediate_operand" " r,m")))]
-  ""
+  "!TARGET_ZBB"
```
kito-cheng, line 1141: It's standard pattern name so we still need that even TARGET_ZBB has a pattern in bitmanip.md.

You can add a define_expand pattern above to resolve such issue
```lisp
(define_insn_and_split "extend<SHORT:mode><SUPERQI:mode>2"
  [(set (match_operand:SUPERQI   0 "register_operand"     "")
	(sign_extend:SUPERQI
	    (match_operand:SHORT 1 "nonimmediate_operand" "")))]
```
and then add a `*` to the name of this pattern 
```lisp
(define_insn_and_split "*extend<SHORT:mode><SUPERQI:mode>2"
```

rdolbeau: Should be handled by the second commit

rdolbeau:
@kito-cheng From [this announcement](https://gcc.gnu.org/pipermail/gcc/2021-June/236182.html), it seems the issue of the copyright assignment solved itself :-)

I've added the appropriate sign-off to the PR.

gcc/testsuite/gcc.target/riscv/rvb-zbb-sext_h.c
```diff
@@ -0,0 +1,15 @@
...
+
+#if __riscv_xlen == 32
+int sext_h(int a) {
+	return (short)a;
+}
+#else
+long sext_h(long a) {
+	return (short)a;
+}
+#endif
```
kito-cheng, line 13: I think both using long is fine, since sizeof(long) == sizeof(int) on RV32.

rdolbeau:
added the missing sign-of on the testsuite commits...

rdolbeau:
@kito-cheng Forgot there was two instructions, also added a test for sext.b; this should be complete now.

kito-cheng:
Verified:)

PR #269, opened 2021-06-01
Fix macro _RVVFLOATLD_FF.

kuanlinchentw:
Hi,

As tittle. Just fixed a minor problem in _RVVFLOATLD_FF.


kito-cheng:
Oh, thanks!

PR #271, opened 2021-06-03
Fix rev8.  Is zbb not zbp.

jim-wilson:
Requires making bswapsi2 an expander as rv64 rev8.w is zbp and rv32 rev8
is zbb so we need a non-trivial test to determine if we support bswapsi2.
Also use correct opcodes now that binutils supports them.


jim-wilson:
Was already reviewed and tested by Kito on another tree so merging.

PR #272, opened 2021-06-04
Merge bswapsi2_32 and bswapsi2_64

kito-cheng:
#271 surprisingly break the non-B target, because using define_expand will cause bswap optimizaion always enabled even
 no bswap on target, that will made __bswapsi2 optimize to an infinite  recursion call.

Testcase:
```c
typedef unsigned SItype;
    SItype
__bswapsi2 (SItype u)
{
  return ((((u) & 0xff000000) >> 24)
          | (((u) & 0x00ff0000) >>  8)
          | (((u) & 0x0000ff00) <<  8)
          | (((u) & 0x000000ff) << 24));
}
```

### DIFF
--- a/gcc/config/riscv/bitmanip.md
+++ b/gcc/config/riscv/bitmanip.md
@@ -349,27 +349,16 @@
 
 ;;; ??? grev
 
-(define_expand "bswapsi2"
-  [(set (match_operand:SI 0 "register_operand")
-	(bswap:SI (match_operand:SI 1 "register_operand")))]
-  ""
+(define_insn "bswapsi2"
+  [(set (match_operand:SI 0 "register_operand" "=r")
+        (bswap:SI (match_operand:SI 1 "register_operand" "r")))]
+  "TARGET_ZBB || TARGET_ZBP"
 {
-  if (!(TARGET_ZBP || (TARGET_ZBB && !TARGET_64BIT)))
-    FAIL;
-})
-
-(define_insn "bswapsi2_32"
-  [(set (match_operand:SI 0 "register_operand" "=r")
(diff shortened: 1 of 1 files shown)


jim-wilson:
You are right.  My patch doesn't work in the non-B case.

You patch incidentally is missing the constraints since you took the define_expand which has no constraints instead of the define_insn which does.

There is one case where my patch does better, in the rv64_zbb case for your testcase, my patch gives
__bswapsi2:
	rev8	a0,a0
	srai	a0,a0,32
	ret
because the failure of bswapsi2 causes it to back off and use bswapdi2 which does exist.  Your patch gives about a dozen instructions with shifts and ors.

Thinking about this some more, I think we should just make bswapsi2 emit the rev8/srai sequence for rv64gc_zbb.  There can optionally be a splitter, but we can worry about that later.  For now it is simpler without the splitter.  This makes the condition for bswapsi2 the same as bswapdi2 without the TARGET_64BIT check.  So I'm suggesting instead

(define_insn "bswapsi2"
  [(set (match_operand:SI 0 "register_operand" "=r")
        (bswap:SI (match_operand:SI 1 "register_operand" "r")))]
  "TARGET_ZBB || TARGET_ZBP"
{
  if (TARGET_64BIT)
    return TARGET_ZBB ? "rev8\t%0,%1\n\tsrai\t%0,%0" : "greviw\t%0,%1,0x18";
  else
    return "rev8\t%0,%1";
}

kito-cheng:
Thanks, updated!

Changes:
- Add constraint
- Fix rv64*_zbb.
- Use rev8.w rather than greviw with 0x18.

jim-wilson (approved):

PR #276, opened 2021-06-18
make regression fixes for a memcpy testcase.

jim-wilson:
Cherry picking patches from FSF GCC to get the fix.


kito-cheng:
Thanks :)

PR #285, opened 2021-08-09
Add vector mask load/store functions.

rjiejie:


rjiejie:
@kito-cheng 
Review this patch please :)

rjiejie:
@kito-cheng Hello ? :)

gcc/config/riscv/vector.md
```diff
@@ -5933,6 +5933,58 @@
...
+	  [(unspec:VMASKMODES
+	    [(mem:VMASKMODES (match_operand 1 "register_operand"))]
+	    UNSPEC_VLM)
+	   (reg:SI VL_REGNUM)]
+	 UNSPEC_USEVL))]
+  "TARGET_VECTOR"
+{
+})
+
+(define_insn "*lm<mode>"
```
kito-cheng, line 5950: You don't need to separated `define_insn`/`define_expand` pattern if they have same RTL pattern, only `define_insn` is fine.

gcc/config/riscv/riscv_vector.h
```diff
@@ -3487,6 +3487,20 @@ v##NAME##_f_f_w_f##SEW##m##LMUL##_m (					\
 _RVV_WFLOAT_ITERATOR_ARG (_RVV_FLOAT_NCVT_FF, nfcvt_ff, fncvt)
 _RVV_WFLOAT_ITERATOR_ARG (_RVV_FLOAT_NCVT_FF, nfcvt_rod_ff, fncvt_rod)
 
+#define _RVV_MASK_LOAD_STORE(MLEN, OP)					\
+__extension__ extern __inline vbool##MLEN##_t				\
+__attribute__ ((__always_inline__, __gnu_inline__, __artificial__))	\
+vlm_v_b##MLEN (const uint8_t *base, word_type vl)			\
+{									\
+  return __builtin_riscv_vlmbool##MLEN (base);				\
```
kito-cheng, line 3495: `vlm`/`vsm` still need to set `VL` by `vsetvl*`, but I saw all other mask operation didn't handle that correctly too, I guess we need to fix that later, so I am OK without setting `VL` for now.

gcc/config/riscv/vector.md
rjiejie: They are not same completely, the expand need to translate address (register type) to memory type.

kito-cheng: Oh, I got it

PR #286, opened 2021-08-17
Bugfix changed vl error

rjiejie:
From vector spec https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#avl-encoding
we know that "vsetvli x0,x0,*" allows the vtype register to be changed while maintaining the current vl, provided VLMAX is not reduced, but "mov" SPN can not guarantee this constraint in some context, so we need to call vsetvlmax explicitly.

### DIFF
--- a/gcc/config/riscv/vector.md
+++ b/gcc/config/riscv/vector.md
@@ -274,6 +274,16 @@
   [(set_attr "type" "vector")
    (set_attr "mode" "none")])
 
+(define_insn "vsetvlmax<VIMODES:mode>_<P:mode>"
+  [(set (reg:VIMODES VTYPE_REGNUM) (const_int UNSPECV_VSETVL))
+   (set (reg:P VL_REGNUM) (const_int UNSPECV_VSETVL))
+   (set (match_operand:P 0 "register_operand" "=r")
+	(reg:P VL_REGNUM))]
+  "TARGET_VECTOR"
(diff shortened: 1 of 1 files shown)


rjiejie:

rjiejie:
Make vector spec issue also, see https://github.com/riscv/riscv-v-spec/issues/722

gcc/config/riscv/vector.md
```diff
@@ -274,6 +274,16 @@
...
    (set_attr "mode" "none")])
 
+(define_insn "vsetvlmax<VIMODES:mode>_<P:mode>"
+  [(set (reg:VIMODES VTYPE_REGNUM) (const_int UNSPECV_VSETVL))
+   (set (reg:P VL_REGNUM) (const_int UNSPECV_VSETVL))
+   (set (match_operand:P 0 "register_operand" "=r")
+	(reg:P VL_REGNUM))]
+  "TARGET_VECTOR"
+  "vsetvli\t%0,x0,e<VIMODES:sew>,m<VIMODES:lmul>"
+  [(set_attr "type" "vset")
```
kito-cheng, line 284: `type` should be `vector`, I guess that `type` used for your internal branch :p

rjiejie: Indeed :)

Fix it

PR #287, opened 2021-08-23
Bugfix for register overlap of vector indexed segment loads

rjiejie:
see spec
https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#783-vector-indexed-segment-loads-and-stores


rjiejie:

kito-cheng:
Could you also fix that for `"*vseg_idx_<order>load<VTMODES:mode><VIMODES:mode>_<P:mode>_mask"`, masked version should have same issue?

rjiejie:
> Could you also fix that for `"*vseg_idx_<order>load<VTMODES:mode><VIMODES:mode>_<P:mode>_mask"`, masked version should have same issue?

Fixed :)

PR #290, opened 2021-08-30
Bugfix riscv_gen_load_poly_int error of flmul

rjiejie:


rjiejie:

kito-cheng:
Did you have testcase to demonstrate the bug?

gcc/config/riscv/riscv.c
```diff
@@ -4271,47 +4271,55 @@ riscv_gen_load_poly_int (rtx target, rtx tmp1, rtx tmp2, poly_int64 value)
...
+      else
+	{
+	  emit_insn (insn);
+	  value -= UNITS_PER_V_REG * vlenb_mul_int;
+	}
     }
-  else
+  if (tail)
     {
-      gcc_assert (multiple_p (value, UNITS_PER_V_REG));
```
rjiejie, line 4300: Remove assert because that the FLMUL exist and not multiple of UNITS_PER_V_REG

gcc/config/riscv/riscv.c
```diff
@@ -4271,47 +4271,55 @@ riscv_gen_load_poly_int (rtx target, rtx tmp1, rtx tmp2, poly_int64 value)
...
+  bool tail = true;
 
   if (value.coeffs[0] != value.coeffs[1])
     scalar_offset = value.coeffs[0] - value.coeffs[1];
 
   if (scalar_offset)
     value -= scalar_offset;
 
-  if (maybe_lt (value, BYTES_PER_RVV_VECTOR))
+  if (!maybe_lt (known_lt (value, 0) ? -value : value, UNITS_PER_V_REG))
```
rjiejie, line 4282: Stack value is negative, so we need to check with zero,
and then compared with UNITS_PER_V_REG

rjiejie:
> Did you have testcase to demonstrate the bug?

I did not have it by now, is it obvious from my comments ?
I think so :)

rjiejie:
@kito-cheng Hello ?

rjiejie:
@kito-cheng Any comments for this issue ?

PR #292, opened 2021-08-31
[T-HEAD][RISCV] Refactor some declarations and fix name mangling errors

pcwang-thead:
Use macro `_RVV_INT_ITERATOR` and `_RVV_FLOAT_ITERATOR` to reduce
redundant declarations and fix name mangling errors for RVV types.


pcwang-thead:
The name of builtin-types after name mangling should start with the length of type name, or `c++filt` program can't recognize these symbols.
For example:
```
$ c++filt _Z5paramI_vfloat32m1_tEvT_
_Z5paramI_vfloat32m1_tEvT_

$ c++filt _Z5paramI13_vfloat32m1_tEvT_
void param<_vfloat32m1_t>(_vfloat32m1_t)
```

PR #293, opened 2021-08-31
Add range checking for vget/vset

pcwang-thead:
Add checker to `riscv_builtin_description` so that we can do some static semantic checking for builtins during expanding.

Currently, range checking for vget/vset have been implemented.

There may have a drawback of this implementation: we can't get exact source location of caller after inlining, which are real common in `riscv_vector.h`.

```
In file included from rvv-vget-vset.c:3:
riscv_vector.h: In function 'vget':
riscv_vector.h:264:1: error: index 2 is out of range, should be less than 2.
  264 | _RVV_INT_WLMUL_ITERATOR (_RVVINT_VECTOR_EXTRACT)
      | ^~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/gcc/config/riscv/riscv-builtins.c
+++ b/gcc/config/riscv/riscv-builtins.c
@@ -38,6 +38,7 @@ along with GCC; see the file COPYING3.  If not see
 #include "langhooks.h"
 #include "stringpool.h"
 #include "riscv-vector-iterator.h"
+#include "riscv-protos.h"
 
 /* We don't want the PTR definition from ansi-decl.h.  */
 #undef PTR
@@ -129,6 +130,17 @@ enum riscv_builtin_type {
    return (COND);			\
  }
 
+#define DECL_CHECKER(NAME) \
+static bool                \
+check_##NAME (tree, rtx, rtx, machine_mode);
+
+#define DEF_CHECKER(NAME)                        \
+static bool                                      \
+check_##NAME (tree exp ATTRIBUTE_UNUSED,         \
+              rtx target ATTRIBUTE_UNUSED,       \
+              rtx subtarget ATTRIBUTE_UNUSED,    \
+              machine_mode mode ATTRIBUTE_UNUSED)
+
 /* This structure describes a single built-in function.  */
 struct riscv_builtin_description {
   /* The code of the main .md file instruction.  See riscv_builtin_type
@@ -146,10 +158,18 @@ struct riscv_builtin_description {
 
   /* Whether the function is available.  */
   unsigned int (*avail) (void);
+
+  /* Checker for the function.  */
+  bool (*check) (tree, rtx, rtx, machine_mode);
 };
 
 AVAIL (hard_float, TARGET_HARD_FLOAT)
 AVAIL (vector, TARGET_VECTOR)
+
+DECL_CHECKER(vector_insert)
+DECL_CHECKER(vector_tuple_insert)
+DECL_CHECKER(vector_extract)
+DECL_CHECKER(vector_tuple_extract)
 
 /* Construct a riscv_builtin_description from the given arguments.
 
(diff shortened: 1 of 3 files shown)


pcwang-thead:
@kito-cheng Any comments?

gcc/testsuite/gcc.target/riscv/rvv/rvv-vget-vset.c
```diff
@@ -0,0 +1,27 @@
+/* { dg-do compile } */
+
+#include <riscv_vector.h>
+
+vint32m1_t vget(vint32m2_t data)
+{
+  return vget_v_i32m2_i32m1(data, 2);
+  /* { dg-error {index 2 is out of range, should be less than 2.} } */
```
kito-cheng, line 8: I got failed for this case, I guess it's because the error line is mis-matched:
```
/scratch1/kitoc/riscv-gnu-workspace/rv64gcv-public/build/build-gcc-newlib-stage2/gcc/include/riscv_vector.h: In function 'vget':^M
/scratch1/kitoc/riscv-gnu-workspace/rv64gcv-public/build/build-gcc-newlib-stage2/gcc/include/riscv_vector.h:264:1: error: index 2 is out of range, should be less than 2.^M
```

And I can pass that after I update the check line to:
```
 /* { dg-error "index 2 is out of range, should be less than 2." "" { target *-*-* } 0 } */
```

gcc/testsuite/gcc.target/riscv/rvv/rvv-vget-vset.c
```diff
@@ -0,0 +1,27 @@
...
+}
+
+vint32m1_t vget_tuple(vint32m1x4_t data)
+{
+  return vget_i32m1x4_i32m1(data, 4);
+  /* { dg-error {index 4 is out of range, should be less than 4.} } */
```
kito-cheng, line 14: Ditto.

gcc/testsuite/gcc.target/riscv/rvv/rvv-vget-vset.c
```diff
@@ -0,0 +1,27 @@
...
+}
+
+vint32m8_t vset(vint32m8_t data, vint32m1_t v)
+{
+  return vset_v_i32m1_i32m8(data, 8, v);
+  /* { dg-error {index 8 is out of range, should be less than 8.} } */
```
kito-cheng, line 20: Ditto.

gcc/testsuite/gcc.target/riscv/rvv/rvv-vget-vset.c
```diff
@@ -0,0 +1,27 @@
...
+}
+
+vint32m1x8_t vset_tuple(vint32m1x8_t data, vint32m1_t v)
+{
+  return vset_i32m1x8(data, 8, v);
+  /* { dg-error {index 8 is out of range, should be less than 8.} } */
```
kito-cheng, line 26: Still got failed even change this line to `/* { dg-error "index 8 is out of range, should be less than 8." "" { target *-*-* } 0 } */` 
I guess it because it's same as last `dg-error` check, but seems like change the testcase to 9 instead of 8 can resolve the issue.

gcc/config/riscv/riscv-builtins.c
```diff
@@ -3346,3 +3405,69 @@ riscv_atomic_assign_expand_fenv (tree *hold, tree *clear, tree *update)
   *clear = build_call_expr (fsflags, 1, old_flags);
   *update = NULL_TREE;
 }
+
+static bool
+check_range(location_t loc, tree tuple, tree index, machine_mode emode)
```
kito-cheng, line 3410: Need space between function name and '('.

kito-cheng (commented):
Only few minor issue, otherwise LGTM :)

pcwang-thead:
@kito-cheng I have resolved all the issues, is it OK now?

kito-cheng:
Thanks!

PR #294, opened 2021-09-01
Add a target hook to remove noop moves

pcwang-thead:


kito-cheng:
I would suggest that we should fix the move pattern rather than add hook for this case, `mov` isn't really use `VL_REGNUM`, so I think we can made the move pattern like scalar move, just `(set (reg) (reg))`, the reason why we use `VL_REGNUM` their is because those implementation are written before whole register move/load/store added to the ISA.

```
(define_insn "*mov<mode>"
  [(set (match_operand:VMODES 0 "reg_or_mem_operand"  "=vr,vr,m")
        (unspec:VMODES
          [(match_operand:VMODES 1 "reg_or_mem_operand"  "vr,m,vr")
           (reg:SI VL_REGNUM)]
         UNSPEC_USEVL))]
  "TARGET_VECTOR"
  "@
   vmv<xlmul>r.v\t%0,%1
   vl<xlmul>r.v\t%0,%1
   vs<xlmul>r.v\t%1,%0"
  [(set_attr "type" "vector")
   (set_attr "mode" "none")])

```

pcwang-thead:
> 
> 
> I would suggest that we should fix the move pattern rather than add hook for this case, `mov` isn't really use `VL_REGNUM`, so I think we can made the move pattern like scalar move, just `(set (reg) (reg))`, the reason why we use `VL_REGNUM` their is because those implementation are written before whole register move/load/store added to the ISA.
> 
> ```
> (define_insn "*mov<mode>"
>   [(set (match_operand:VMODES 0 "reg_or_mem_operand"  "=vr,vr,m")
>         (unspec:VMODES
>           [(match_operand:VMODES 1 "reg_or_mem_operand"  "vr,m,vr")
>            (reg:SI VL_REGNUM)]
>          UNSPEC_USEVL))]
>   "TARGET_VECTOR"
>   "@
>    vmv<xlmul>r.v\t%0,%1
>    vl<xlmul>r.v\t%0,%1
>    vs<xlmul>r.v\t%1,%0"
>   [(set_attr "type" "vector")
>    (set_attr "mode" "none")])
> ```

OK, I will have a try.

PR #299, opened 2021-09-15
Add vector floating-point absolute value functions

rjiejie:
reference to https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/master/intrinsic_funcs/07_vector_floating-point_functions.md


rjiejie:
@kito-cheng any comments ?

kito-cheng:
Thanks :)

PR #301, opened 2021-09-17
Change factor of DWARF_POLY_INDETERMINATE_VALUE to 16

pcwang-thead:
Actually I haven't figure it out clearly how it works.

In my view: the value of `factor` in SVE is 2, and `VG` in SVE is the number of 64-bit elements in an SVE vector; similarly, `vlenb` is the number of bytes (8-bit elements) in an RVV vector, the `factor` should be `16 = 2 * 64 / 8`.

At least, the generated DWARF expressions match up with assembly code if factor is 16.
For example:
```c
#include <stdio.h>
#include <stddef.h>
#include <riscv_vector.h>

uint8_t a[16] = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16};
uint8_t b[16] = {};

int main()
{

    size_t gvl = vsetvlmax_e8m1();
    vuint8m1_t v = vle8_v_u8m1(a, gvl);
    vse8_v_u8m1(b, v, gvl);

    for (size_t i = 0; i < 16; i++)
    {
        printf("%d ", b[i]);
    }
    printf("\n");
    
    return 0;
}
```
The code to calculate `v`'s offset is:
```asm
	.loc 1 12 20
	csrr	a4,vlenb
	li	a5,-1
	mul	a5,a4,a5
	addi	a5,a5,-16
	addi	a4,s0,-16
	add	a5,a4,a5
	vs1r.v	v1,0(a5)
```
The DWARF expression of `v` are like:
```dwarf
 # when factor is `8`
(DW_OP_fbreg: -16; DW_OP_bregx: 7202 (vlenb) 0; DW_OP_const1s: -2; DW_OP_mul; DW_OP_lit16; DW_OP_minus; DW_OP_plus)
                                                                ^ 
# when factor is `16`
(DW_OP_fbreg: -16; DW_OP_bregx: 7202 (vlenb) 0; DW_OP_const1s: -1; DW_OP_mul; DW_OP_lit16; DW_OP_minus; DW_OP_plus)
                                                                ^ 
```

### DIFF
--- a/gcc/config/riscv/riscv.c
+++ b/gcc/config/riscv/riscv.c
@@ -5875,10 +5875,10 @@ static unsigned int
 riscv_dwarf_poly_indeterminate_value (unsigned int i, unsigned int *factor,
 				      int *offset)
 {
-  /* Polynomial invariant 1 == (VLENB / 8) - 1.  */
+  /* Polynomial invariant 1 == (VLENB / 16) - 1.  */
   /* XXX: It's might not correct for ELEN=32 system.  */
   gcc_assert (i == 1);
-  *factor = 8;
+  *factor = 16;
   *offset = 1;
   return RISCV_DWARF_VLEN;
 }


pcwang-thead:
@kito-cheng Any comments?

PR #302, opened 2021-09-17
Bugfix memory load/store error of unit-stride Zvlsseg

rjiejie:


rjiejie:
@kito-cheng any comments ?

PR #303, opened 2021-09-17
Bugfix vsetvl error of vector integer extension

rjiejie:
reference to https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#113-vector-integer-extension


rjiejie:
@kito-cheng any comments ?

PR #304, opened 2021-09-22
Support of Zpsfoperand inline assembly

marcfedorow:


jim-wilson:
GCC patches require either a copyright assignment or a valid DCO Signed-off-by line.  I don't see either.  See
[https://gcc.gnu.org/contribute.html](url)

The P extension work is being done by PLCT.  This needs to be merged with their work.

We only accept development work upstream in general.  We only accept patches for frozen/ratified extensions on mainline, and drafts can go in the integration/experimental branches.  Though some extension work is still happening on branches in riscv-gcc, which includes the P work.

You submitted this for the gcc-10 mainline branch which isn't appropriate.  Not clear why.  There is no explanation of what this is for.  Maybe you meant to send this elsewhere?

pz9115:
You can try to send this patch into https://github.com/linsinan1995/riscv-gcc/tree/riscv-gcc-experiment-p-ext, after we check it works, it will be merged, thanks.

jim-wilson:
If you accept patches without proper FSF copyright assignment or DCO sign off then we will not be able to upstream the P ext work and it will have to be thrown away.  Talk to Wei Wu about copyright assignments.

pz9115:
Sorry for I misunderstand the mean and thank you for th remind, we will follow agreement strictly with patches have proper FSF copyright assignment or DCO sign off and refuse any others, I just mean it been discuss to implement in correct place.

We have a check with this patch and find that Lin had already implement the odd/even regs limits in riscv.c,
https://github.com/linsinan1995/riscv-gcc/commit/45bdc5c300bc74f168d3ec457621fe288cb0841c
so maybe we can close this PR.

PR #305, opened 2021-09-23
RISC-V: mulm3_highpart -> smulm3_highpart

Geng-Qi-alibaba:
The correct standard pattern name is 'smulm3_highpart'

### DIFF
--- a/gcc/config/riscv/riscv.md
+++ b/gcc/config/riscv/riscv.md
@@ -649,14 +649,14 @@
   emit_insn (gen_muldi3 (low, operands[1], operands[2]));
 
   rtx high = gen_reg_rtx (DImode);
-  emit_insn (gen_<u>muldi3_highpart (high, operands[1], operands[2]));
+  emit_insn (gen_<su>muldi3_highpart (high, operands[1], operands[2]));
 
   emit_move_insn (gen_lowpart (DImode, operands[0]), low);
   emit_move_insn (gen_highpart (DImode, operands[0]), high);
   DONE;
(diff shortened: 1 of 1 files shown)


kito-cheng:
Thanks, generally LGTM, could you include a testcase for that?

cmuellner:
@kito-cheng, shouldn't something like this go into mainline first?

jim-wilson:
Yes, upstream first.  And why patch gcc-10?  We already moved to gcc-11.

jim-wilson:
Current sources have a mulv<mode>4 pattern that also needs fixing, as it calls gen_mul<mode>3_highpart.

Not sure how practical it is to find a testcase unless you already have one.  There are very few places that call the highpart multiply functions directly without going through the widening multiply patterns, and the windening multiply patterns are using the non-canonical names, so they do work.

Geng-Qi:
> Thanks, generally LGTM, could you include a testcase for that?

Sorry for late reply.

I discovered this issue by accident, so I didn't have a testcase at the time.
I have recently tried various methods to try to build a testcase, but unfortunately in the end I was unable to make it.

As there is no testcase, I am not sure myself that this fix actually works.

Geng-Qi:
Following all the advice, I will commit it to gcc-patches.

kito-cheng:
This patch ahs merged into trunk, so I gonna close this, and I won't back port this to `riscv-gcc-10.*` or r`iscv-gcc-11.*` branch since this won't effect the code gen.

PR #307, opened 2021-09-29
Add Non-Tuple API for indexed of Zvlsseg

rjiejie:


rjiejie:
@kito-cheng Have any free time to review this patch ?

PR #310, opened 2021-10-14
Fix VTYPE_REGNUM without v-extension

rjiejie:


rjiejie:
This patch conserve the VTYPE register when no V-ext to ensure that ira is not disturbed.

@kito-cheng Any comments ?

PR #311, opened 2021-10-20
Fix vfwredsum_vs* to vfwredusum_vs* du to vector spec v1.0

rjiejie:
See https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/master/rvv-intrinsic-api.md#154-vector-widening-floating-point-reduction-operations


rjiejie:
@kito-cheng Ping ...

PR #317, opened 2021-12-03
Fix vfredsum_vs* to vfredusum_vs* du to vector spec v1.0

rjiejie:
Similar as https://github.com/riscv-collab/riscv-gcc/pull/311

See https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#143-vector-single-width-floating-point-reduction-instructions


rjiejie:
@kito-cheng Any comments ?

PR #327, opened 2022-02-24
analyzer: Fix tests for glibc 2.35 [PR101081]

palmer-dabbelt:
I haven't actually tested this one yet, but it very much smells like it'll fix the new failures I'm running into when trying to update glibc.  I'll comment further when the tests finish, but I've got to go run the beer hour (which is more like 3 hours) now so I figured I'd open this preemptively.


palmer-dabbelt:
Yep, it's back to the single failure

PR #376, opened 2023-04-20
RISCV: Add vector psabi checking.

yanzhang-dev:
This patch adds support to check function's argument or return is vector type and throw warning.

gcc/ChangeLog:

	* config/riscv/riscv.cc (riscv_legitimize_move): (riscv_vector_psabi_warnning): (riscv_arg_has_vector): (riscv_pass_in_vector_p): (riscv_get_arg_info):

gcc/testsuite/ChangeLog:

	* gcc.target/riscv/vector-abi-1.c: New test.
	* gcc.target/riscv/vector-abi-2.c: New test.

### DIFF
--- a/gcc/config/riscv/riscv.cc
+++ b/gcc/config/riscv/riscv.cc
@@ -3790,8 +3790,9 @@ riscv_pass_in_vector_p (const_tree type)
 
   if (type && riscv_arg_has_vector (type) && !warned)
     {
-      warning (OPT_Wpsabi, "ABI for the vector type is currently in experimental"
-	       "stage and may changes in the upcoming version of GCC.");
+      warning (OPT_Wpsabi, "ABI for the vector type is currently in "
+	       "experimental stage and may changes in the upcoming version of "
+	       "GCC.");
       warned = 1;
     }
 }
--- a/gcc/config/riscv/riscv.cc
+++ b/gcc/config/riscv/riscv.cc
@@ -3736,10 +3736,10 @@ riscv_scalable_vector_type_p (const_tree type)
 {
   tree size = TYPE_SIZE (type);
   if (size && TREE_CODE (size) == INTEGER_CST)
-    return true;
+    return false;
 
   /* For the data type like vint32m1_t, the size code is POLY_INT_CST.  */
-  return false;
+  return true;
 }
 
 static bool
@@ -3762,7 +3762,7 @@ riscv_arg_has_vector (const_tree type)
 
 	    /* Ignore it if it's fixed length vector.  */
 	    if (VECTOR_TYPE_P (field_type))
-	      is_vector = !riscv_scalable_vector_type_p (field_type);
+	      is_vector = riscv_scalable_vector_type_p (field_type);
 	    else
 	      is_vector = riscv_arg_has_vector (field_type);
 	  }
@@ -3770,7 +3770,7 @@ riscv_arg_has_vector (const_tree type)
       break;
 
     case VECTOR_TYPE:
-      is_vector = !riscv_scalable_vector_type_p (type);
+      is_vector = riscv_scalable_vector_type_p (type);
       break;
 
     default:
@@ -3791,7 +3791,7 @@ riscv_pass_in_vector_p (const_tree type)
 
   if (type && riscv_arg_has_vector (type) && !warned)
     {
-      warning (OPT_Wpsabi, "ABI for the vector type is currently in "
+      warning (OPT_Wpsabi, "ABI for the scalable vector type is currently in "
 	       "experimental stage and may changes in the upcoming version of "
 	       "GCC.");
       warned = 1;
--- a/gcc/config/riscv/riscv.cc
+++ b/gcc/config/riscv/riscv.cc
@@ -3798,6 +3798,28 @@ riscv_pass_in_vector_p (const_tree type)
     }
 }
 
+/* Initialize a variable CUM of type CUMULATIVE_ARGS
+   for a call to a function whose data type is FNTYPE.
+   For a library call, FNTYPE is 0.  */
+
+void
+init_cumulative_args (CUMULATIVE_ARGS *cum,  /* Argument info to initialize */
+		      tree fntype,	/* tree ptr for function decl */
+		      rtx libname,	/* SYMBOL_REF of library name or 0 */
+		      tree fndecl,
+		      int caller)
+{
+  memset (cum, 0, sizeof (*cum));
+
+  if (fndecl)
+    {
+      const tree_function_decl &fn = FUNCTION_DECL_CHECK (fndecl)->function_decl;
+
+      if (fn.built_in_class == NOT_BUILT_IN)
+	  cum->rvv_psabi_warning = 1;
+    }
+}
+
 /* Fill INFO with information about a single argument, and return an
    RTL pattern to pass or return the argument.  CUM is the cumulative
    state for earlier arguments.  MODE is the mode of this argument and
@@ -3882,8 +3904,11 @@ riscv_get_arg_info (struct riscv_arg_info *info, const CUMULATIVE_ARGS *cum,
 	}
     }
 
-  /* Only check existing of vector type.  */
-  riscv_pass_in_vector_p (type);
+  if (cum->rvv_psabi_warning)
+    {
+      /* Only check existing of vector type.  */
+      riscv_pass_in_vector_p (type);
+    }
 
   /* Work out the size of the argument.  */
   num_bytes = type ? int_size_in_bytes (type) : GET_MODE_SIZE (mode).to_constant ();
@@ -3972,7 +3997,18 @@ riscv_function_value (const_tree type, const_tree func, machine_mode mode)
     }
 
   memset (&args, 0, sizeof args);
-  return riscv_get_arg_info (&info, &args, mode, type, true, true);
+
+  const_tree arg_type = type;
+  if (func && DECL_RESULT (func))
+    {
+      const tree_function_decl &fn = FUNCTION_DECL_CHECK(func)->function_decl;
+      if (fn.built_in_class == NOT_BUILT_IN)
+	args.rvv_psabi_warning = 1;
+
+      arg_type = TREE_TYPE (DECL_RESULT (func));
+    }
+
+  return riscv_get_arg_info (&info, &args, mode, arg_type, true, true);
 }
 
 /* Implement TARGET_PASS_BY_REFERENCE. */
--- a/gcc/config/riscv/riscv.cc
+++ b/gcc/config/riscv/riscv.cc
@@ -3728,11 +3728,18 @@ riscv_pass_fpr_pair (machine_mode mode, unsigned regno1,
 				   GEN_INT (offset2))));
 }
 
-static void
-riscv_vector_psabi_warnning ()
+/* Use the TYPE_SIZE to distinguish the type with vector_size attribute and
+   intrinsic vector type.  Because we can't get the decl for the params.  */
+
+static bool
+riscv_arg_has_vector_size_attribute(const_tree type)
 {
-  warning (OPT_Wpsabi, "ABI for the vector type is currently in experimental"
-       "stage and may changes in the upcoming version of GCC.");
+  tree size = TYPE_SIZE(type);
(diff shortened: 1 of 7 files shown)


gcc/testsuite/gcc.target/riscv/vector-abi-2.c
```diff
@@ -0,0 +1,14 @@
+/* { dg-do compile } */
+/* { dg-options "-march=rv64gcv -mabi=lp64d" } */
+
+#include "riscv_vector.h"
+
+vint32m1_t
+fun (vint32m1_t* a) {  return *a; }  /* { dg-warning "the vector type" } */
```
kito-cheng, line 7: Could you add a test that pass or return pointer of vector type only, and make sure their won't have warning on that case?

gcc/config/riscv/riscv.cc
```diff
@@ -2089,8 +2089,8 @@ riscv_legitimize_move (machine_mode mode, rtx dest, rtx src)
 	}
       return true;
     }
-  /* Expand 
-       (set (reg:QI target) (mem:QI (address))) 
+  /* Expand
+       (set (reg:QI target) (mem:QI (address)))
```
kito-cheng: Don't fix this in this patch

gcc/config/riscv/riscv.cc
```diff
@@ -2105,7 +2105,7 @@ riscv_legitimize_move (machine_mode mode, rtx dest, rtx src)
 
       temp_reg = gen_reg_rtx (word_mode);
       zero_extend_p = (LOAD_EXTEND_OP (mode) == ZERO_EXTEND);
-      emit_insn (gen_extend_insn (temp_reg, src, word_mode, mode, 
+      emit_insn (gen_extend_insn (temp_reg, src, word_mode, mode,
```
kito-cheng: Don't fix this in this patch

gcc/config/riscv/riscv.cc
```diff
@@ -3728,6 +3728,65 @@ riscv_pass_fpr_pair (machine_mode mode, unsigned regno1,
...
+
+	    if (VECTOR_TYPE_P (type))
+	      is_vector = true;
+	    else
+	      is_vector = riscv_arg_has_vector (TREE_TYPE (f));
+	  }
+
+      break;
+
+    case VECTOR_TYPE:
```
kito-cheng, line 3772: Does it included fixed length vector type like `typedef int v4si __attribute__ ((vector_size (16)));`?

gcc/testsuite/gcc.target/riscv/vector-abi-2.c
yanzhang-dev: Yes. Added a new test case with `dg-bogus` to check no warning threw. Is that OK?

gcc/config/riscv/riscv.cc
yanzhang-dev: DONE.

gcc/config/riscv/riscv.cc
yanzhang-dev: Done.

gcc/config/riscv/riscv.cc
yanzhang-dev: Yes. It will treat the variable with this attribute as a vector. I think it's what we want. Is it?

zhongjuzhe:
Not familiar with ABI. I think it's all kito's call :)

gcc/testsuite/gcc.target/riscv/vector-abi-2.c
kito-cheng: Ok

gcc/config/riscv/riscv.cc
kito-cheng: That should exclude since it not passed in vector reg

yanzhang-dev: Hi Kito. I want to filter out this case when the type has attribute `vector_size`. But found `vector_size` is for variable, which means we can't get it by `TYPE_ATTRIBUTES`. And haven't found the way to convert the `type` to `decl`. Do you have any idea to distinguish them ?

kito-cheng: Type size?

yanzhang-dev: Thanks Kito. Tested this API and found for `vint32m_1`, it will return `POLY_INT_CST` and for the other will return `INTEGER_CST`. It can work now with new test cases.

gcc/config/riscv/riscv.cc
```diff
@@ -3728,6 +3728,74 @@ riscv_pass_fpr_pair (machine_mode mode, unsigned regno1,
...
+
+static void
+riscv_pass_in_vector_p (const_tree type)
+{
+  static int warned = 0;
+
+  if (type && riscv_arg_has_vector (type) && !warned)
+    {
+      warning (OPT_Wpsabi, "ABI for the vector type is currently in experimental"
+	       "stage and may changes in the upcoming version of GCC.");
```
Incarnation-p-lee: will the message become something like `in experimentalstage and ...` instead of `in experimental stage and ...` after joining?

gcc/config/riscv/riscv.cc
```diff
@@ -3728,6 +3728,74 @@ riscv_pass_fpr_pair (machine_mode mode, unsigned regno1,
...
+  return is_vector;
+}
+
+/* Pass the type to check whether it's a vector type or contains vector type.
+   Only check the value type and no checking for vector pointer type.  */
+
+static void
+riscv_pass_in_vector_p (const_tree type)
+{
+  static int warned = 0;
```
Incarnation-p-lee, line 3790: not sure if can avoid the local static variable here or something like `bool*` can get similar logic implemented.

gcc/config/riscv/riscv.cc
yanzhang-dev: Yes. You're right. Will modify it in the fix commit.

gcc/config/riscv/riscv.cc
yanzhang-dev: I found it seems like this is a uniform pattern for warning checking. To align with them, such as fpr abi checking, I use the same logic here. Another point is, the `bool*` needs a global var to store, how to name and manage so many global warning/error variables can be headache.

gcc/config/riscv/riscv.cc
```diff
@@ -3728,6 +3728,75 @@ riscv_pass_fpr_pair (machine_mode mode, unsigned regno1,
 				   GEN_INT (offset2))));
 }
 
+/* Use the TYPE_SIZE to distinguish the type with vector_size attribute and
+   intrinsic vector type.  Because we can't get the decl for the params.  */
+
+static bool
+riscv_arg_has_vector_size_attribute (const_tree type)
```
kito-cheng: `riscv_scalable_vector_type_p`

gcc/config/riscv/riscv.cc
```diff
@@ -3728,6 +3728,75 @@ riscv_pass_fpr_pair (machine_mode mode, unsigned regno1,
...
+      if (!COMPLETE_TYPE_P (type))
+	break;
+
+      for (tree f = TYPE_FIELDS (type); f; f = DECL_CHAIN (f))
+	if (TREE_CODE (f) == FIELD_DECL)
+	  {
+	    if (!TYPE_P (TREE_TYPE (f)))
+	      break;
+
+	    /* If there's vector_size attribute, ignore it.  */
```
kito-cheng: `Ignore it if it's fixed length vector.`

gcc/config/riscv/riscv.cc
```diff
@@ -3728,6 +3728,75 @@ riscv_pass_fpr_pair (machine_mode mode, unsigned regno1,
...
+	    if (VECTOR_TYPE_P (TREE_TYPE (f)))
+	      is_vector = !riscv_arg_has_vector_size_attribute (type);
```
kito-cheng: `riscv_arg_has_vector_size_attribute (TREE_TYPE (f))`?

gcc/config/riscv/riscv.cc
Incarnation-p-lee: Cool, thank you!

gcc/config/riscv/riscv.cc
Incarnation-p-lee: I see, it is good enough to follow the convention.

gcc/config/riscv/riscv.cc
yanzhang-dev: Done.

gcc/config/riscv/riscv.cc
yanzhang-dev: Done.

gcc/config/riscv/riscv.cc
yanzhang-dev: Thanks, you're right. It should be the field type.

gcc/config/riscv/riscv.cc
```diff
@@ -3728,6 +3728,76 @@ riscv_pass_fpr_pair (machine_mode mode, unsigned regno1,
...
+
+static bool
+riscv_scalable_vector_type_p (const_tree type)
+{
+  tree size = TYPE_SIZE (type);
+  if (size && TREE_CODE (size) == INTEGER_CST)
+    return true;
+
+  /* For the data type like vint32m1_t, the size code is POLY_INT_CST.  */
+  return false;
```
kito-cheng: The return value should be inverted after function rename :)

gcc/config/riscv/riscv.cc
```diff
@@ -3728,6 +3728,76 @@ riscv_pass_fpr_pair (machine_mode mode, unsigned regno1,
...
+   Only check the value type and no checking for vector pointer type.  */
+
+static void
+riscv_pass_in_vector_p (const_tree type)
+{
+  static int warned = 0;
+
+  if (type && riscv_arg_has_vector (type) && !warned)
+    {
+      warning (OPT_Wpsabi, "ABI for the vector type is currently in "
```
kito-cheng: `vector` -> `scalable vector`, that should more clear that it's exclude fixed length vector.

gcc/config/riscv/riscv.cc
yanzhang-dev: Oh understand. Could I know why the intrinsic vector type is named as `scalable vector` ?

kito-cheng: that's the term defined in AArch64 SVE actually, and we follow to use that term in our backend too , e.g. riscv.cc:riscv_v_adjust_scalable_frame

yanzhang-dev:
Hi @kito-cheng , The commit [51fbc2e](https://github.com/riscv-collab/riscv-gcc/pull/376/commits/51fbc2ebbffef1cb7c4556e5752492922b342acf) tries to fix the warning on intrinsic APIs. I'm not sure whether it's a right solution ?

I think the problem is we want to distinguish below 3 scenarios,

1. `vint32m1_t fun(xxx)` -> warning on return type.
2. `vint32m1_t a = fun(xxx)` -> no warning
3. `xxx fun(vint32m1_t)` -> warning on arg.

Both three will call `riscv_pass_by_reference` and then call `riscv_get_arg_info`. So we can't put the checking in `get_arg_info`. My solution moves the checking  into `function_arg` and `function_value`. It can successfully distinguish the three.

But there's another new issue. When compiling with `-flto`, the warning will disappear and will show when do the linking. So I skip the the lto in tests.

gcc/config/riscv/riscv.h
```diff
@@ -671,14 +671,18 @@ typedef struct {
...
   unsigned int num_fprs;
+
+  bool rvv_psabi_warning;
 } CUMULATIVE_ARGS;
 
 /* Initialize a variable CUM of type CUMULATIVE_ARGS
    for a call to a function whose data type is FNTYPE.
    For a library call, FNTYPE is 0.  */
 
+void init_cumulative_args (CUMULATIVE_ARGS *, tree, rtx, tree, int);
```
kito-cheng, line 682: Rename to `riscv_init_cumulative_args` and put into `riscv-protos.h`.

kito-cheng (commented):
Only one minor comment, you can go to mailing list after you address that :)

